PR #10, opened 2025-12-11
Remove unneeded Gitlab Folder

younes-makhchan:
The .gitlab folder is no longer needed, as we are using downstream pipelines


.gitlab-ci.yml
```diff
@@ -1,9 +1,5 @@
 image : $GITLAB_CI_CD_IMAGE
```
ElayachiFouad: To be changed to JQ image

X-SP33D:
```suggestion
image : $GITLAB_CI_CD_IMAGE_JQ
```

PR #26, opened 2025-12-17
Add support for flows in CrewAI adapter

gojkoc54:
### AgentSpec -> CrewAI

- supported: StartNode, EndNode, ToolNode, LlmNode, InputMessageNode, OutputMessageNode
- not supported: branching (most notably)

### CrewAI -> AgentSpec

- all nodes are mapped to ToolNode
- not supported: branching (routers, listeners with multiple triggers)

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
+++ b/pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
@@ -5,6 +5,7 @@
 # (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
 
 
+import inspect
 import uuid
 from typing import (
     Any,
@@ -12,6 +13,7 @@ from typing import (
     Dict,
     List,
     Optional,
+    Tuple,
     Type,
     Union,
     cast,
@@ -25,12 +27,20 @@ from pydantic import BaseModel
 from pyagentspec.adapters.crewai._types import (
     CrewAIAgent,
     CrewAIBaseTool,
+    CrewAIFlow,
     CrewAILlm,
     CrewAIStructuredTool,
     CrewAITool,
+    FlowState,
 )
 from pyagentspec.agent import Agent as AgentSpecAgent
 from pyagentspec.component import Component as AgentSpecComponent
+from pyagentspec.flows.edges import ControlFlowEdge, DataFlowEdge
+from pyagentspec.flows.flow import Flow as AgentSpecFlow
+from pyagentspec.flows.node import Node as AgentSpecNode
+from pyagentspec.flows.nodes import EndNode as AgentSpecEndNode
+from pyagentspec.flows.nodes import StartNode as AgentSpecStartNode
+from pyagentspec.flows.nodes import ToolNode as AgentSpecToolNode
 from pyagentspec.llms import LlmConfig as AgentSpecLlmConfig
 from pyagentspec.llms import LlmGenerationConfig as AgentSpecLlmGenerationConfig
 from pyagentspec.llms.ollamaconfig import OllamaConfig as AgentSpecOllamaModel
@@ -121,6 +131,10 @@ class CrewAIToAgentSpecConverter:
             )
         elif isinstance(crewai_component, CrewAIBaseTool):
             agentspec_component = self._tool_convert_to_agentspec(
+                crewai_component, referenced_objects
+            )
+        elif isinstance(crewai_component, CrewAIFlow):
+            agentspec_component = self._flow_convert_to_agentspec(
                 crewai_component, referenced_objects
             )
         else:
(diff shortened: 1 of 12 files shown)


pyagentspec/tests/adapters/crewai/flows/test_llmnode_crewai.py
```diff
@@ -0,0 +1,67 @@
...
+        ],
+        outputs=[car_property],
+    )
+
+    flow_instance = AgentSpecLoader().load_component(flow)
+    result = flow_instance.kickoff({"nationality": "italian"})
+
+    assert isinstance(result, dict)
+    assert car_property.title in result
+    assert "ital" in result[car_property.title].lower()
```
sonleoracle: This might be a flaky test, can you simply assert that the content is not empty

pyagentspec/tests/adapters/crewai/flows/test_llmnode_crewai.py
```diff
@@ -0,0 +1,67 @@
...
+                source_node=llm_node,
+                source_output=car_property.title,
+                destination_node=end_node,
+                destination_input=car_property.title,
+            ),
+        ],
+        outputs=[car_property],
+    )
+
+    flow_instance = AgentSpecLoader().load_component(flow)
```
sonleoracle, line 68: Could you assert this is a CrewAI object?

pyagentspec/tests/adapters/crewai/conftest.py
```diff
@@ -54,3 +54,14 @@ def fake_response(self, method: str, endpoint: str, **kwargs) -> requests.Respon
...
+
+@pytest.fixture
+def mute_crewai_event_bus(monkeypatch):
+    try:
+        from crewai.events.event_bus import crewai_event_bus
+    except Exception:
+        return
+
+    # Replace emit/aemit with no-ops to avoid background threads, rich output, and side effects
+    monkeypatch.setattr(crewai_event_bus, "emit", lambda *args, **kwargs: None, raising=True)
```
sonleoracle: I don't think we should do this, since we will support tracing for the crewai adapter, which relies on the crewai event bus. If you want to mute crewai console prints, you could try:

```python
from crewai.events.event_listener import event_listener as default_listener
from crewai.events.utils.console_formatter import ConsoleFormatter
default_listener.formatter = ConsoleFormatter(verbose=False)
```

pyagentspec/tests/adapters/crewai/test_crewai_to_agentspec.py
```diff
@@ -0,0 +1,100 @@
...
+
+        @start()
+        def generate_sentence_count(self) -> int:
+            return choice([1, 2, 3])
+
+        @listen(generate_sentence_count)
+        def generate_poem(self, sentence_count) -> str:
+            try:
+                llm = CrewAILlm(
+                    model="openai//storage/models/Llama-3.1-70B-Instruct",
```
sonleoracle: why does it need to be prepended with `openai//` ?

pyagentspec/tests/adapters/crewai/test_crewai_to_agentspec.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.fixture
+def agentspec_exporter() -> "AgentSpecExporter":
+    from pyagentspec.adapters.crewai import AgentSpecExporter
+
+    return AgentSpecExporter()
+
+
+@pytest.fixture
+def crewai_flow() -> "CrewAIFlow":
+    from crewai import LLM as CrewAILlm
```
sonleoracle, line 32: In addition to the Crew AI LLM, would you think it makes sense to have a flow node which calls litellm.completion? Like here https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/crew-ai/python/ag_ui_crewai/examples/agentic_chat.py

(this allows native tool calls instead of crewai ReACT prompting)

pyagentspec/tests/adapters/crewai/test_crewai_to_agentspec.py
```diff
@@ -0,0 +1,100 @@
...
+
+        @listen(generate_poem)
+        def add_title(self, poem_text) -> str:
+            return f"The Best Poem\n{poem_text}"
+
+    return PoemFlow()
+
+
+def _get_tool_registry(flow: "CrewAIFlow") -> Dict[str, Any]:
+    return flow._methods
```
sonleoracle, line 66: Could you put a quick comment explaining what this is (looks like it's an internal attribute because of the leading underscore)?

pyagentspec/tests/adapters/crewai/flows/test_inputmessagenode_crewai.py
```diff
@@ -0,0 +1,55 @@
...
+from pyagentspec.flows.nodes import EndNode, InputMessageNode, StartNode
+from pyagentspec.property import StringProperty
+
+
+@pytest.mark.usefixtures("mute_crewai_event_bus")
+def test_inputmessagenode_can_be_imported_and_executed(monkeypatch) -> None:
+    from pyagentspec.adapters.crewai import AgentSpecLoader
+
+    # Mock interactive input() used by InputMessageNodeExecutor
+    monkeypatch.setattr("builtins.input", lambda: "3")
```
sonleoracle, line 22: Interesting idea, maybe we can test the client tool conversion with this too (since it just prompts for user `input`)?

pyagentspec/src/pyagentspec/adapters/crewai/_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    }
+    if json_schema["type"] == "object":
+        # We could do better in inferring the type of values, for now we just use Any
+        return "Dict[str, Any]"
+    if json_schema["type"] == "array":
+        return f"List[{_json_schema_type_to_python_annotation(json_schema['items'])}]"
+    return mapping.get(json_schema["type"], "Any")
+
+
+def create_pydantic_model_from_properties(
```
sonleoracle: If you copied this from the langgraph adapter, please check it again, we pushed some code to make it support nested attributes

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,268 @@ def _agent_convert_to_agentspec(
...
+            )
+
+            tool = AgentSpecServerTool(
+                name=method_name,
+                description=f"Converted CrewAI flow method '{method_name}'",
+                inputs=node_inputs,
+                outputs=node_outputs,
+            )
+            node = AgentSpecToolNode(
+                name=str(method_name),
```
sonleoracle: Is method_name not a string?

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,268 @@ def _agent_convert_to_agentspec(
...
+        # Connect END to all end methods
+        for end_method in end_methods:
+            control_flow_edges, data_flow_edges = self._add_end_edges(
+                nodes[end_method],
+                end_node,
+                control_flow_edges,
+                data_flow_edges,
+            )
+
+        flow_name = getattr(crewai_flow, "name", None) or crewai_flow.__class__.__name__
```
sonleoracle: the flow does not always have a `.name` attribute?

sonleoracle (approved):
LGTM!

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,267 @@ def _agent_convert_to_agentspec(
...
+            ]
+
+            # Create output properties for nodes:
+            # - if there is a return value annotation, add a single "object" property
+            return_annotation = signature.return_annotation
+            has_output = not (
+                return_annotation is inspect.Signature.empty or return_annotation is None
+            )
+            node_outputs = (
+                [AgentSpecProperty(title=f"{method_name}_output", json_schema={"type": "object"})]
```
cesarebernardis: Is the return type really an `object`, or could it be a "base type" like integer or so? Note that `object` in json schema is a mapping (i.e., basically a dict).

If you want to have an equivalent of `Any` type here, you can leave the json schema empty: https://json-schema.org/understanding-json-schema/basics. Same for `node_inputs` right before.

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,267 @@ def _agent_convert_to_agentspec(
...
+            tool = AgentSpecServerTool(
+                name=method_name,
+                description=f"Converted CrewAI flow method '{method_name}'",
+                inputs=node_inputs,
+                outputs=node_outputs,
+            )
+            node = AgentSpecToolNode(
+                name=method_name,
+                tool=tool,
+                inputs=node_inputs,
```
cesarebernardis: Nit: the ToolNode will inherit Tool's inputs and outputs, you can skip making them explicit

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,267 @@ def _agent_convert_to_agentspec(
...
+        )
+        nodes[start_node.name] = start_node
+        referenced_objects[start_node.name] = start_node
+
+        control_flow_edges: list[ControlFlowEdge] = []
+        data_flow_edges: list[DataFlowEdge] = []
+
+        # Connect START to all start methods
+        for start_method in getattr(crewai_flow, "_start_methods", []):
+            control_flow_edges, data_flow_edges = self._add_start_edges(
```
cesarebernardis: This is an invalid configuration in Agent Spec, because a node's branch can only have a single control flow edge connected, while here we might connect multiple edges to the same branch of the start node.

The solution depends on what is the expected behavior in case multiple start nodes are defined in CrewAI: is execution started in parallel? How is the joining of the execution of separate, concurrent branches handled?

To support this you should either serialize everything, or to build separate `ParallelFlowNode`s handling the parallelism properly, but it might be a bit complicated for now. I think that for this first implementation you can just raise a not-supported-error in case multiple start nodes are defined, and we'll tackle the issue in another PR. You can also create an issue to keep track of this.

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,267 @@ def _agent_convert_to_agentspec(
...
+        start_method_names = set(getattr(crewai_flow, "_start_methods", []))
+        for method_name in methods_by_name.keys():
+            method_callable = methods_by_name[method_name]
+
+            # Get method signature to infer properties since they're not explicitly defined
+            signature = inspect.signature(method_callable)
+
+            # Create input properties for nodes:
+            # - for start nodes we use their input parameters
+            # - for other nodes we either leave it empty or have a single "object" property
```
cesarebernardis: It would be nice to explain why. In general, I think this method is missing an explanation of the choices made to accomplish the translation, it would be nice to add them here and there where these comments are placed.

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
```diff
@@ -217,3 +231,267 @@ def _agent_convert_to_agentspec(
...
+        control_flow_edges.append(
+            ControlFlowEdge(
+                name=f"{trigger_node.name}_to_{listener_node.name}_control_edge",
+                from_node=trigger_node,
+                to_node=listener_node,
+            )
+        )
+        trigger_node_outputs = trigger_node.outputs or []
+        listener_node_inputs = listener_node.inputs or []
+        if len(trigger_node_outputs) == 1 and len(listener_node_inputs) == 1:
```
cesarebernardis, line 421: I am not sure I get why this is done only if those nodes have a single output. Is it because the whole state is being carried over? If that's the reason, please add a comment, otherwise we need to connect all the inputs/outputs.

pyagentspec/src/pyagentspec/adapters/crewai/_crewaiconverter.py
```diff
@@ -150,6 +125,155 @@ def convert(
...
 
+    def _flow_convert_to_crewai(
+        self,
+        flow: AgentSpecFlow,
+        tool_registry: Dict[str, CrewAIServerToolType],
+        converted_components: Dict[str, Any],
+    ) -> CrewAIFlow[FlowState]:
+
+        node_executors: Dict[str, NodeExecutor] = {
+            node.id: self._node_convert_to_crewai(
```
cesarebernardis: It's usually better to call `convert` in these cases, as it checks if the node was already converted and returns it without re-performing the conversion.

pyagentspec/src/pyagentspec/adapters/crewai/_template_rendering.py
```diff
@@ -0,0 +1,33 @@
...
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+import re
+from typing import Any, Dict, List, Tuple
+
+from pyagentspec.templating import TEMPLATE_PLACEHOLDER_REGEXP
+
+
+def render_template(template: str, inputs: Dict[str, Any]) -> str:
```
cesarebernardis: These method are now offered by `pyagentspec.adapters._utils`

pyagentspec/src/pyagentspec/adapters/crewai/_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    if isinstance(t, list):
+        variants = [
+            _build_type_from_schema(f"{name}Alt{i}", {"type": subtype}, registry)
+            for i, subtype in enumerate(t)
+        ]
+        return Union[tuple(variants)]
+
+    # arrays
+    if t == "array":
+        items_schema = schema.get("items", {"type": "any"})
```
cesarebernardis: `{"type": "any"}` I don't think is accepted in json schema, please check.

But I guess you added it to the mapping at the end, so it's going to work.

pyagentspec/tests/adapters/crewai/flows/test_llmnode_crewai.py
```diff
@@ -0,0 +1,77 @@
...
+def test_llmnode_can_be_imported_and_executed() -> None:
+    from crewai import Flow as CrewAIFlow
+
+    from pyagentspec.adapters.crewai import AgentSpecLoader
+
+    nationality_property = StringProperty(title="nationality")
+    car_property = StringProperty(title="car")
+    llm_config = OpenAiCompatibleConfig(
+        name="llm_config",
+        model_id="/storage/models/Llama-3.1-70B-Instruct",
```
cesarebernardis: Please use Llama 3.3 or OSS instead, same in the other tests

pyagentspec/tests/adapters/crewai/flows/test_llmnode_crewai.py
```diff
@@ -0,0 +1,77 @@
...
+        url=os.environ.get("LLAMA_API_URL"),
```
cesarebernardis, line 29: We have a variable in the adapters `conftest.py` file, you can import that

pyagentspec/tests/adapters/crewai/test_crewai_to_agentspec.py
gojkoc54: Without this it raises an error:

```
litellm.exceptions.BadRequestError: litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=/storage/models/Llama-3.3-70B-Instruct
 Pass model as E.g. For 'Huggingface' inference endpoints pass in `completion(model='huggingface/starcoder',..)` Learn more: https://docs.litellm.ai/docs/providers
```

pyagentspec/tests/adapters/crewai/flows/test_llmnode_crewai.py
gojkoc54: That one doesn't work with crewai for some reason, the openaicompatible one works

pyagentspec/src/pyagentspec/adapters/crewai/_utils.py
gojkoc54: I stole these functions from the langgraph adapter, seems like you added this piece of code 

https://github.com/oracle/agent-spec/blame/main/pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py#L165-L178

Anyways, maybe makes more sense to move these to pyagentspec/adapters/_utils.py

pyagentspec/src/pyagentspec/adapters/crewai/_agentspecconverter.py
gojkoc54: From all the docs I've seen and some simple tests I did, seems like crewai flows only support single inputs and outputs in their nodes/methods. They probably expect you to use dicts/lists/... if you want to achieve this. I'll add some exceptions here with proper error messages

Example:
```
from crewai import Flow as CrewAIFlow
from crewai.flow.flow import listen, start

class TestFlow(CrewAIFlow):
    @start()
    def f1(self) -> int:
        return 1, 2

    @listen(f1)
    def f2(self, a, b) -> int:
        return a + b

flow = TestFlow()

result = flow.kickoff()
```

This raises:
`TypeError: TestFlow.f2() missing 1 required positional argument: 'b'`

PR #28, opened 2025-12-18
sec: temporarily remove crewai from adapters

paul-cayet:
## Context

The CrewAI adapter is temporarily removed from the Agent Spec adapters until the security vulnerability is addressed on the Crew AI side.

PR #30, opened 2025-12-18
docs: update docs with temporary adapter removal

paul-cayet:
## Context

Docs updates after https://github.com/oracle/agent-spec/pull/28/

### DIFF
--- a/development/_static/landing.css
+++ b/development/_static/landing.css
@@ -111,7 +111,6 @@ p.detail-title {
 }
 
 a {
-  display: inline !important;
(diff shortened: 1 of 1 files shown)


development/_static/landing.css
```diff
@@ -111,6 +111,7 @@ p.detail-title {
 }
 
 a {
+  display: inline !important;
```
ElayachiFouad: To be removed

ElayachiFouad (approved):

PR #34, opened 2025-12-18
Docs - Language spec page redirect

cesarebernardis:
Update installation instructions. Update redirect of language spec page based on docs version.


development/_downloads/f19b8bcca917257b409d935580c6e8a9/constraints.txt
```diff
@@ -2,6 +2,7 @@ jsonschema==4.23.0
 pydantic==2.12.4
 pyyaml==6.0.3
 httpx==0.28.1
+urllib3==2.6.2
```
ElayachiFouad: is this a third party dependency?

dhilloulinoracle: no, 4th party of langchain and other adapters

ElayachiFouad: i believe it should be removed from the constraints file and keep it only in the setup.py since we define only the third party dependencies (that we have BA for) in the constraints file

cesarebernardis: We will include also 4th party dependencies in this constraints file, so this line can stay

PR #38, opened 2025-12-22
feat: Datastores and MessageTransforms in agent-spec

youssef62:
**Context**
 We want to have summarization transforms (`ConversationSummarizationTransform` and `MessageSummarizationTransform`) as native agent-spec components. For that we need to bring datastores to agent-spec `OracleDatabaseDatastore`, `PostgresDatastore` and `InMemoryDatastore`).

**Changes**

- Added components: `OracleDatabaseDatastore`, `TlsOracleDatabaseConnectionConfig`, `MTlsOracleDatabaseConnectionConfig`, `PostgresDatabaseDatastore`, `TlsPostgresDatabaseConnectionConfig`, `InMemoryCollectionDatastore`, `MessageSummarizationTransform`, and `ConversationSummarizationTransform`.
- Added tests in relevant files under `tests/serialization/datastores/` and `tests/serialization/transforms/`.
- Fixed a deserialization issue that prevented use of Optional fields.

addresses https://github.com/oracle/agent-spec/issues/50 and https://github.com/oracle/agent-spec/issues/49


docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1947,164 @@ The ManagerWorkers has two main parameters:
   - Workers cannot interact with the end user directly.
   - When invoked, each worker can leverage its equipped tools to complete the assigned task and report the result back to the group manager.
 
+
+Datastores
+~~~~~~~~~~
+
+The core abstract is Datastore which enables agentic systems to store, and access data.
```
lfaucon: I don't really understand this sentence "The core abstract is Datastore ..."

Do you mean: "Datastores are the AgentSpec concept that enables ..."

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1947,164 @@ The ManagerWorkers has two main parameters:
...
+- ``sslcert``: Path of the client SSL certificate, replacing the default `~/.postgresql/postgresql.crt`. Ignored if an SSL connection is not made.
+- ``sslkey``: Path of the file containing the secret key used for the client certificate, replacing the default `~/.postgresql/postgresql.key`. Ignored if an SSL connection is not made.
+- ``sslrootcert``: Path of the file containing SSL certificate authority (CA) certificate(s). Used to verify server identity.
+- ``sslcrl``: Path of the SSL server certificate revocation list (CRL). Certificates listed will be rejected while attempting to authenticate the server's certificate.
+
+
+Transforms
+^^^^^^^^^^
+
+Transforms extend the base ``MessageTransform`` component can be used in agents. they apply a transformation to the messages before it's passed to the LLM.
```
lfaucon:
```suggestion
Transforms extend the base ``MessageTransform`` component can be used in agents. They apply a transformation to the messages before they are passed to the LLM.
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1947,164 @@ The ManagerWorkers has two main parameters:
...
+
+.. code-block:: python
+
+    class MessageTransform(Component, abstract=True):
+        pass
+
+MessageSummarizationTransform
+'''''''''''''''''''''''''''''
+
+Summarizes messages exceeding a given number of characters using an LLM and caches the summaries in a ``Datastore``. This is useful for conversations with long messagwes where the context can become too large for the agent LLM to handle.
```
lfaucon:
```suggestion
Summarizes messages exceeding a given number of characters using an LLM and caches the summaries in a ``Datastore``. This is useful for conversations with long messages where the context can become too large for the agent LLM to handle.
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1947,164 @@ The ManagerWorkers has two main parameters:
...
+        min_num_messages: int = 10
+        summarization_instructions: str = (
+            "Please make a summary of this conversation. Include relevant information and keep it short. "
+            "Your response will replace the messages, so just output the summary directly, no introduction needed."
+        )
+        summarized_conversation_template: str = "Summarized conversation: {{summary}}"
+        datastore: Optional[Datastore] = None
+        max_cache_size: Optional[int] = 10_000
+        max_cache_lifetime: Optional[int] = 4 * 3600
+        cache_collection_name: str = "summarized_conversations_cache"
```
lfaucon: I think it's useful to extend a bit the specification document, to give some more details about the expected execution behaviour for each of the parameters of the two message transforma configuration classes.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
lfaucon, line 1: Need to update the section `Agent` in this file too

docs/pyagentspec/source/changelog.rst
```diff
@@ -89,6 +89,16 @@ New features
   Introduced ManagerWorkers in the Agent Spec Language Specification.
   For more information check out the corresponding :doc:`managerworkers how-to guide <howtoguides/howto_managerworkers>` or read the :ref:`API Reference <managerworkers>`.
 
+* **Datastores**
+
+  Added support for datastores in Agent Spec through :ref:`OracleDatabaseDatastore <oracledatabasedatastore>` and :ref:`PostgresDatabaseDatastore <postgresdatabasedatastore>`.
+  Datastores enable persistent storage and caching capabilities for agent workflows. :ref:`InMemoryCollectionDatastore <inmemorycollectiondatastore>` provides a drop-in replacement for development and prototyping.
+
+* **Transforms**
```
lfaucon:
```suggestion
* **Context Summarization Transforms**
```

So that it's more obvious for developers reading the the changelog what the new thing is.

docs/pyagentspec/source/changelog.rst
lfaucon, line 1: We will have to move the changelog entries to be within the correct release

pyagentspec/src/pyagentspec/datastores/datastore.py
```diff
@@ -0,0 +1,50 @@
...
+    raise ValueError(
+        "Property is not a valid entity property. Entity properties must be objects with properties."
+    )
+
+
+Entity = Annotated[Property, AfterValidator(is_object_property)]
+
+
+class Datastore(Component, abstract=True):
+    """Base class for Datastores. Datastores store and retrieve data."""
```
lfaucon: minor:
```suggestion
    """Base class for Datastores. Datastores configure how to store and retrieve data."""
```

pyagentspec/tests/agentspec_configs/example_tools_with_confirmation_25_4_2.yaml
lfaucon, line 1: The changes to all the test example configs will probably not occur when we set the min version of transforms to be `> 25.4.2`

lfaucon (commented):
Nice work 💯 Thanks

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: I clarified this to: 
> Datastores are the AgentSpec abstraction that enables agentic systems to store, and access data.

in 8242b83.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Corrected in 8242b83.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Thanks for catching this. Corrected in 8242b83.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: I agree, I added comments explaining each parameter in d480c8f.

docs/pyagentspec/source/changelog.rst
youssef62: Good idea. Added to 8242b83.

pyagentspec/src/pyagentspec/datastores/datastore.py
youssef62: Added in 8242b83.

pyagentspec/tests/agentspec_configs/example_tools_with_confirmation_25_4_2.yaml
youssef62: Yes, in [c29c4eb](https://github.com/oracle/agent-spec/pull/38/commits/c29c4ebcb35ad4b40e9a75f5282d3f95ac4e0471) I made these feature belong to version `26.1.1`.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Good catch, thank you. Added in [c144d54](https://github.com/oracle/agent-spec/pull/38/commits/c144d54a4a933422ef153f67d293f952f82d186c).

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -304,6 +282,8 @@ def _load_field(
                 and inspect.isclass(annotation)
                 and issubclass(annotation, Property)
             ):
+                if isinstance(content, annotation):
```
paul-cayet, line 292: Please add a comment for when this is reached

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -381,27 +361,26 @@ def _load_field(
...
                     return self._load_field(content, inner_annotation)
-                except Exception:  # nosec
+                except Exception as e:  # nosec
                     # Something went wrong in deserialization,
-                    # it's not the right type, we try the next one
-                    pass
-
-            # We tried all the components and pydantic models, and it did not work out,
-            # only python type is left. If it is only normal python types, just return the content
-            if self._is_python_type(annotation):
```
paul-cayet, line 406: Why is this gone?

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -381,27 +361,26 @@ def _load_field(
                     return None, []
                 inner_annotations.remove(type(None))
 
+                # This check is needed because `is_optional_type` accepts Union[T,G,None]
```
paul-cayet: Is that necessary given that the accumulated errors are already raised below?

pyagentspec/src/pyagentspec/agent.py
```diff
@@ -54,6 +55,8 @@ class Agent(AgenticComponent):
     """List of toolboxes that are passed to the agent."""
     human_in_the_loop: bool = True
     """Flag that determines if the Agent can request input from the user."""
+    transforms: List[MessageTransform] = []
+    """Additional message transforms that the agent will use to modify its vision of the messages"""
```
geeooleea: Would the agent use the transform, or is the executor applying the transformations before giving the messages to the agent itself?

pyagentspec/src/pyagentspec/transforms/summarization.py
```diff
@@ -0,0 +1,77 @@
+# Copyright © 2025 Oracle and/or its affiliates.
+#
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+"""This file defines message transforms for message and conversation summarization."""
```
geeooleea, line 6: Do we typically use module-level docstring? I don't think I saw any other file using these

pyagentspec/src/pyagentspec/transforms/summarization.py
```diff
@@ -0,0 +1,77 @@
...
+        "Your response will replace the message, so just output the summary directly, no introduction needed."
+    )
+    """Instructions for the LLM on how to summarize messages."""
+
+    summarized_message_template: str = "Summarized message: {{summary}}"
+    """Template for formatting the summarized message output."""
+
+    datastore: Optional[Datastore] = None
+    """
+    Datastore on which to store the cache. If None, an in-memory Datastore will be created automatically.
```
geeooleea: Based on the implementation here, it doesn't seem like the in-memory datastore would be created automatically, I am afraid? I say this because it seems to me like it is up to the executor to choose which datastore to use (if any). The more natural behaviour, based on how caching typically works, would be that nothing is cached and transformations are re-applied every time. A better solution IMO could be to enforce it at the Spec level, by setting the default always to InMemoryCollectionDatastore if it is None.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1951,164 @@ The ManagerWorkers has two main parameters:
...
+        pass
+
+MessageSummarizationTransform
+'''''''''''''''''''''''''''''
+
+Summarizes messages exceeding a given number of characters using an LLM and caches the summaries in a ``Datastore``. This is useful for conversations with long messages where the context can become too large for the agent LLM to handle.
+
+.. code-block:: python
+
+    class MessageSummarizationTransform(MessageTransform):
```
geeooleea, line 2084: Are these acutal defaults, or a usage example? Typically, I think we have actual definitions here, but I'm not sure prompts should be part of it?

geeooleea: Rather, I think the general feeling in AgentSpec is that parameters seldom have defaults. @cesarebernardis might have more info/opinions on it IIRC

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1951,164 @@ The ManagerWorkers has two main parameters:
...
+    class ConversationSummarizationTransform(MessageTransform):
+        llm: LlmConfig  # LLM to use for the summarization.
+        max_num_messages: int = 50  # Number of message after which we trigger summarization.
+        min_num_messages: int = 10  # Number of recent messages to keep from summarizing.
+        summarization_instructions: str = (
+            "Please make a summary of this conversation. Include relevant information and keep it short. "
+            "Your response will replace the messages, so just output the summary directly, no introduction needed."
+        )  # Instruction for the LLM on how to summarize the conversation.
+        summarized_conversation_template: str = "Summarized conversation: {{summary}}"  # Jinja2 template on how to present the summary (with variable `summary`) to the agent using the transform.
+        datastore: Optional[Datastore] = None  # Datastore on which to store the cache. If None, an in-memory Datastore will be created automatically.
```
geeooleea: Referring to my other comment, about the default value of the datastore: it should probably be part of the actual text in the spec, that transformed messages MUST be cached if a runtime wants to be agent-spec compliant, and that if the user doesn't provide a value, then the caching MUST happen in-memory. Then I could get past the fact that datastore is a truly optional parameter that doesn't default to in-memory.

EDIT: New to reviewing on github, but this is the comment that is blocking my approval: IMO we need to be a bit clearer on the behaviour of these (specifically wrt to reproducibility across different calls to the same transform).

pyagentspec/src/pyagentspec/versioning.py
```diff
@@ -30,7 +30,8 @@ class AgentSpecVersionEnum(Enum):
     v25_4_1 = "25.4.1"
     v25_4_2 = "25.4.2"
     v26_1_0 = "26.1.0"
-    current_version = "26.1.0"
+    v26_1_1 = "26.1.1"
+    current_version = "26.1.1"
```
geeooleea: Should this change be made a specific PR? (The bumping of current_version and the creation of the agentspec JSON?). Side question, shouldn't we have a json_spec nightly to mirror the language spec nightly?

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
lfaucon: Because it was wrong.
Why would a python type be ok to deserialize a Union type that deos not specify that this python type is actually correct?
This actually caused very hard to debug bugs that I investigated with @youssef62.
If there are clear valuable examples where this makes sense, please let us know so that we can add relevant tests and revert the changes.

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
lfaucon: I agree, it seems superfluous. Happy to remove this to have a little bit less code :)

pyagentspec/src/pyagentspec/transforms/summarization.py
lfaucon: In agent-spec we do, yes

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
lfaucon: Good point. I agree with having no defaults in the specs (but keep default in the pyagentspec SDK for developer's convenience)

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
lfaucon: Thanks for the comment.

I actually like the proposal that `None` should mean no caching, but `Not Specified` should be up to the Runtime and PyAgentSpec SDK to fill-in the most helpful default value (probably InMemoryDatastore)

- So in the specs: `datastore: Optional[Datastore]`
- And in the SDK: `datastore: Optional[Datastore] = Field(default_factory=_get_default_in_memory_cache`

-------

whether we go for this or not, let's indeed double check which documentation / instructions belong in the Specs and/or in the SDK

pyagentspec/src/pyagentspec/agent.py
lfaucon: (it's a bit confusing what "agent" means exactly, so I may not understand correctly what you mean)

The runtime must apply all transforms to messages of the ongoing conversation before any prompt is submitted to the Agent's LLM.

pyagentspec/src/pyagentspec/versioning.py
lfaucon: Yes, we are just anticipating this change here to help implement the rest of the changes correctly.

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
youssef62: Hello, this is needed for `pyagentspec/tests/test_partialcomponentconstruction.py::test_llm_node_can_be_constructed_with_missing_inputs` to pass.

Here content=`StringProperty(json_schema={'title': 'a', 'type': 'string'}...`. 

**Before these changes to: `deseriazationcontext.py`**  We used to try to pass `content` as `json_schema` to `Property` in ` Property(json_schema=content)` which would fail. This would also make `Optional[List[StringProperty]]` fail to deserialize. However, this used to pass `if self._is_python_type(annotation):` which results in returning `content,[]`

**After the changes** We directly return `content, []` whenever we detect that an already deserialized property is passed.

I think that in either case, `content` here should never be a deserialized property `deserialized property`. It looks to me that it contradicts the type annotation `BaseModelAsDictT`. So our fix here does not solve the underlying issue.

In the context of this PR, I think solving the issue like this is reasonable, as we left the code functionally equivalent and added a comment for future reference (a74b92a) . Let me know what you think.

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
youssef62: I agree that this is not necessary, thanks for catching that. I removed it in e9a0ec9.

pyagentspec/src/pyagentspec/agent.py
geeooleea: I was commenting on the wording of "the agent will use the transforms to modify its vision on the messages [...]". This makes it sound like the message transform is some kind of tool that is provided to the agent, but that is not the case

youssef62: Thanks @geeooleea. I agree that the comment was a little misleading. I clarified it in 54265f1.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Thank you @geeooleea. I added in what @lfaucon suggested (which also aligns with your other comment)  7c053a54.

pyagentspec/src/pyagentspec/transforms/summarization.py
youssef62: I think this also discussed in your other related comment and should be addressed by 7c053a5.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: I removed the defaults here f4f1ad9.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1951,158 @@ The ManagerWorkers has two main parameters:
...
+
+``TlsOracleDatabaseConnectionConfig`` For standard TLS Oracle Database connections.
+
+.. code-block:: python
+
+    class TlsOracleDatabaseConnectionConfig(OracleDatabaseConnectionConfig):
+        user: SensitiveField[str]
+        password: SensitiveField[str]
+        dsn: SensitiveField[str]
+        protocol: Literal["tcp", "tcps"] = "tcps"
```
geeooleea: In another discussion we agreed to "no defaults in the spec. Should it also apply here?

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1951,158 @@ The ManagerWorkers has two main parameters:
...
+        config_dir: Optional[str] = None
+
+- ``user``: User used to connect to the database
+- ``password``: Password for the provided user
+- ``dsn``: Connection string for the database (e.g., created using `oracledb.make_dsn`)
```
geeooleea: Citing oracledb.make_dsn here is problematic for two reasons:

1. It will only work in python
2. The method has been deprecated since I wrote the wayflow docs

I suggest just removing this. Connection string for the database should be sufficient as there is plenty of documentation around connecting to a db

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1942,6 +1951,158 @@ The ManagerWorkers has two main parameters:
...
+- ``sslcert``: Path of the client SSL certificate, replacing the default `~/.postgresql/postgresql.crt`. Ignored if an SSL connection is not made.
+- ``sslkey``: Path of the file containing the secret key used for the client certificate, replacing the default `~/.postgresql/postgresql.key`. Ignored if an SSL connection is not made.
+- ``sslrootcert``: Path of the file containing SSL certificate authority (CA) certificate(s). Used to verify server identity.
+- ``sslcrl``: Path of the SSL server certificate revocation list (CRL). Certificates listed will be rejected while attempting to authenticate the server's certificate.
+
+
+Transforms
+^^^^^^^^^^
+
+Transforms extend the base ``MessageTransform`` component can be used in agents. They apply a transformation to the messages before they are passed to the agent's LLM.
```
geeooleea:
```suggestion
Transforms extend the base ``MessageTransform`` component that can be used in agents. They apply a transformation to the messages before they are passed to the agent's LLM.
```

youssef62: Thanks for catching that, corrected in 09cb6b5.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: I agree, I removed it in 09cb6b5.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Yes, thanks for catching it. I removed them here 09cb6b5.

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -311,8 +289,10 @@ def _load_field(
                 and inspect.isclass(annotation)
                 and issubclass(annotation, Property)
             ):
-                if isinstance(content, Property):
-                    # already an instantiated property (e.g. partial config)
+                if isinstance(content, annotation):
+                    # This condition should never be reached since content is typed as BaseModelAsDictT.
+                    # However, it is being hit in tests (e.g., test_partialcomponentconstruction.py).
+                    # We need to investigate why deserialized components are passed here.
```
lfaucon:
```suggestion
                    # This condition can be reached when building the component from a partial configuration
                    # which contains Property objects that already built and not represented by only their schema.
```

Let's not write comments about future tasks that we should do.
If really we should investigate this, let's create an issue for it.
I think we don't need to investigate this.
It is the expected behaviour for the partial configurations functionality

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -393,25 +373,20 @@ def _load_field(
 
             # Try to deserialize components/pydantic models according to any of the annotations
             # If any of them works, we will proceed with that. This is our best effort.
+            accumulated_errors = []
             for inner_annotation in inner_annotations:
                 try:
                     return self._load_field(content, inner_annotation)
-                except ValueError:
+                except Exception as e:  # nosec
```
lfaucon: Let's keep the changes that @paul-cayet did in a separate PR, changing this to a ValueError check

```suggestion
                except ValueError as e:
```

pyagentspec/tests/serialization/datastores/test_datastores.py
```diff
@@ -0,0 +1,52 @@
...
+from pyagentspec.serialization.serializer import AgentSpecSerializer
+from pyagentspec.versioning import AgentSpecVersionEnum
+
+from .conftest import DATASTORES_AND_THEIR_SENSITIVE_FIELDS
+
+
+@pytest.mark.parametrize("datastore, sensitive_fields", DATASTORES_AND_THEIR_SENSITIVE_FIELDS)
+def test_can_serialize_and_deserialize_datastore(datastore, sensitive_fields) -> None:
+    serialized_ds = AgentSpecSerializer().to_yaml(datastore)
+    print(serialized_ds)
```
lfaucon: please clean-up print statements

pyagentspec/tests/serialization/transforms/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def create_test_llm_config():
+    config = OpenAiConfig(
+        id="test_llm",
+        name="test_openai_config",
+        model_id="gpt-3.5-turbo",
+    )
+    # Set to v26_1_1 to match the serialized agentspec_version, ensuring deserialized == original
+    config.min_agentspec_version = AgentSpecVersionEnum.v26_1_1
```
lfaucon: minor: Slight preference for doing `assert a._is_equal(b, fields_to_exclude=["min_agentspec_version"])` rather than this

pyagentspec/tests/serialization/test_agent.py
```diff
@@ -273,3 +278,80 @@ def test_deserializing_agent_with_builtin_tools_and_unsupported_version_raises(
...
+
+
+@pytest.mark.parametrize(
+    "datastore, sensitive_fields",
+    DATASTORES_AND_THEIR_SENSITIVE_FIELDS,
+)
+def test_agent_with_non_empty_transforms_can_be_serialized_and_deserialized(
+    datastore, sensitive_fields, vllmconfig
+):
+    print(f"Testing with datastore type: {type(datastore).__name__}")
```
lfaucon: clean-up
```suggestion
```

pyagentspec/tests/serialization/test_agent.py
```diff
@@ -273,3 +278,80 @@ def test_deserializing_agent_with_builtin_tools_and_unsupported_version_raises(
...
+    transforms = [
+        create_message_summarization_transform(datastore),
+        create_conversation_summarization_transform(datastore),
+    ]
+    # The default min_agentspec_version for VllmConfig is v25_4_1. If we leave it unchanged,
+    # the agent with non-empty transforms would serialize to v26_1_1 (due to the transforms requiring that version).
+    # During deserialization, all fields including vllmconfig would be deserialized to v26_1_1,
+    # but vllmconfig's min_agentspec_version would still be v25_4_1, causing the test deserialized == original to fail.
+    # Therefore, we explicitly set the version to v26_1_1
+    vllmconfig.min_agentspec_version = AgentSpecVersionEnum.v26_1_1
```
lfaucon: minor: Slight preference for doing `assert a._is_equal(b, fields_to_exclude=["min_agentspec_version"])` rather than this

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
youssef62: Are we sure that all errors due to being not able to parse are `ValueError`? Couldn't we get `ValidationError` or some other type of error?

pyagentspec/tests/serialization/test_agent.py
youssef62: I agree `vllmconfig.min_agentspec_version = AgentSpecVersionEnum.v26_1_1` is not ideal.

However, to have the equivalent check, we would need to do 
```
    assert deserialized_agent._is_equal(agent, fields_to_exclude=["llm_config"])
    assert deserialized_agent.llm_config._is_equal(agent.llm_config,fields_to_exclude=["min_agentspec_version"] )
```
which seems not very readable. You still prefer it?

pyagentspec/tests/serialization/test_agent.py
youssef62: Done 691930.

pyagentspec/tests/serialization/datastores/test_datastores.py
youssef62: Done in 6919309

pyagentspec/src/pyagentspec/agent.py
```diff
@@ -54,6 +55,8 @@ class Agent(AgenticComponent):
     """List of toolboxes that are passed to the agent."""
     human_in_the_loop: bool = True
     """Flag that determines if the Agent can request input from the user."""
+    transforms: List[MessageTransform] = []
```
jschweiz:
```suggestion
    transforms: List[MessageTransform] = Field(default_factory=list)
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1946,6 +1955,158 @@ The ManagerWorkers has two main parameters:
   - Workers cannot interact with the end user directly.
   - When invoked, each worker can leverage its equipped tools to complete the assigned task and report the result back to the group manager.
 
+
+Datastores
+~~~~~~~~~~
+
+Datastores are the AgentSpec abstraction that enables agentic systems to store, and access data.
```
jschweiz:
```suggestion
Datastores are the AgentSpec abstraction that enables agentic systems to store and access data.
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -382,6 +382,11 @@ following goals :
         discovered from an MCP server.
       - * A ToolBox connecting to an MCP server and exposing its tools to an Agent
 
+    * - Datastore
+      - A component that provides storage capabilities for agentic systems, enabling them to store and access data.
+      - * An in-memory datastore for testing
+        * An Oracle Database datastore for production use
```
jschweiz: missing the postgres one?

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1946,6 +1955,158 @@ The ManagerWorkers has two main parameters:
...
+        pass
+
+All datastores currently defined represent collections of objects defined by their type as an ``Entity``, however the Datastore parent type is left without constraint to enable future extension or the development of plugins with different Datastore structure (e.g. key-value stores, or graph database, or unstructured data storage)
+
+An ``Entity`` is a property that has a JSON schema equivalent to an object property; an object property would work.
+
+Relational datastores & In-Memory datastore
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Relational datastores (``OracleDatabaseDatastore``, ``PostgresDatabaseDatastore``) and their drop-in in-memory replacement (``InMemoryCollectionDatastore``) contain many collections (tables) where each collection is a set entities. They need to have a predefined fixed schema. A schema is a mapping from a collection name to an Entity object defining the entity(row).
```
jschweiz: is there a precise distinction that differenciates `relational` from `non-relational` ones? It's no clear

pyagentspec/src/pyagentspec/transforms/summarization.py
```diff
@@ -0,0 +1,229 @@
...
+    summarization_instructions: str = (
+        "Please make a summary of this message. Include relevant information and keep it short. "
+        "Your response will replace the message, so just output the summary directly, no introduction needed."
+    )
+    """Instructions for the LLM on how to summarize messages."""
+
+    summarized_message_template: str = "Summarized message: {{summary}}"
+    """Template for formatting the summarized message output."""
+
+    max_cache_size: Optional[int] = 10_000
```
jschweiz, line 115: should we add some validation on those fileds like >0, ...

pyagentspec/src/pyagentspec/transforms/summarization.py
```diff
@@ -0,0 +1,229 @@
...
+    ... )
+
+    """
+
+    DEFAULT_COLLECTION_NAME: ClassVar[str] = "summarized_conversations_cache"
+
+    llm: LlmConfig
+    """LLM configuration for conversation summarization."""
+
+    max_num_messages: int = 50
```
jschweiz, line 198: probably we want to ensure that max_num_messages > min_num_message, so we can add a validation?

pyagentspec/src/pyagentspec/datastores/oracle.py
```diff
@@ -0,0 +1,63 @@
+# Copyright © 2025 Oracle and/or its affiliates.
+#
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+"""This module defines the oracle datastore and oracle database connnection configuration components."""
```
jschweiz:
```suggestion
"""This module defines the oracle datastore and oracle database connection configuration components."""
```

pyagentspec/src/pyagentspec/datastores/postgres.py
```diff
@@ -0,0 +1,64 @@
+# Copyright © 2025 Oracle and/or its affiliates.
+#
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+"""This module defines the postgres datastore and postgres database connnection configuration components."""
```
jschweiz:
```suggestion
"""This module defines the postgres datastore and postgres database connection configuration components."""
```

pyagentspec/tests/serialization/datastores/test_datastores.py
```diff
@@ -0,0 +1,51 @@
...
+def test_can_serialize_and_deserialize_datastore(datastore, sensitive_fields) -> None:
+    serialized_ds = AgentSpecSerializer().to_yaml(datastore)
+    assert len(serialized_ds.strip()) > 0
+    deserialized_ds = AgentSpecDeserializer().from_yaml(
+        yaml_content=serialized_ds, components_registry=sensitive_fields
+    )
+    assert deserialized_ds == datastore
+    serialized_ds = AgentSpecSerializer().to_json(datastore)
+    assert len(serialized_ds.strip()) > 0
+    deserialized_ds = AgentSpecDeserializer().from_yaml(
```
jschweiz: isn't it supposed to be from_json here?
```suggestion
    deserialized_ds = AgentSpecDeserializer().from_json(
```

pyagentspec/tests/serialization/transforms/test_summarization_transforms.py
```diff
@@ -0,0 +1,95 @@
...
+    assert len(serialized_transform.strip()) > 0
+
+    deserialized_transform = AgentSpecDeserializer().from_yaml(
+        yaml_content=serialized_transform, components_registry=sensitive_fields
+    )
+    assert deserialized_transform == transform
+
+    serialized_transform = AgentSpecSerializer().to_json(transform)
+    assert len(serialized_transform.strip()) > 0
+    deserialized_transform = AgentSpecDeserializer().from_yaml(
```
jschweiz:
```suggestion
    deserialized_transform = AgentSpecDeserializer().from_json(
```

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
lfaucon: I wondered the same when seeing this change in @paul-cayet PR.

It seems that for all existing components right now, this works as expected with all the quite extensive tests that we have.

Also pydantic's validation errors are covered by this `except`:
```
❯ python 
Python 3.11.1 (main, Jan 19 2023, 11:57:00) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pydantic import ValidationError
>>> issubclass(ValidationError, ValueError)
True
```

Finally, I prefer this code that maybe some day will not catch the error so that we discover problems and fix them, rather than the broad `except Exception` which will hide potential problems from us

pyagentspec/tests/serialization/test_agent.py
lfaucon: Are you sure that's needed to be done with the two assertions?

```
assert deserialized_agent._is_equal(agent, fields_to_exclude=["min_agentspec_version"])
```

The difference between this and your proposal is that we don't check any of the `min_agentspec_version` right?

I find it is a very small difference. I don't have a strong opinion on this.

`vllmconfig.min_agentspec_version = AgentSpecVersionEnum.v26_1_1` also works ok :)
Nice that you added a comment to explain why this surprising thing is included.

pyagentspec/tests/serialization/transforms/test_summarization_transforms.py
lfaucon: Interesting 🤩 The test passes with this "mistake" because JSON syntax is valid yaml (but not the other way around).

pyagentspec/src/pyagentspec/transforms/summarization.py
youssef62: Thank you for the suggestion. I added it here: b42578b.

pyagentspec/src/pyagentspec/transforms/summarization.py
youssef62: Thank you for the suggestion. I added it here: b42578b.

pyagentspec/tests/serialization/test_agent.py
youssef62: > The difference between this and your proposal is that we don't check any of the min_agentspec_version right?

Yes

> I find it is a very small difference. I don't have a strong opinion on this.

Then let's use `_is_equal` which I also prefer. Added in [6aff5d1](https://github.com/oracle/agent-spec/pull/38/commits/6aff5d149c40feaf01b4901ee2049fa2694db97d)

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: I tried to made that clearer here [28cd83a](https://github.com/oracle/agent-spec/pull/38/commits/28cd83a565cf07af157c454fad7709c190fbfed9). Let me know if it's clear enough.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
youssef62: Good catch. Added here [3e038ea](https://github.com/oracle/agent-spec/pull/38/commits/3e038ea142177f1ea65c4f1d8abbdc3378f7c188)

pyagentspec/tests/serialization/transforms/conftest.py
youssef62: Done here [6aff5d1](https://github.com/oracle/agent-spec/pull/38/commits/6aff5d149c40feaf01b4901ee2049fa2694db97d).

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
youssef62: Thank you for the explanation. It's convincing. I added that here: [762bd54](https://github.com/oracle/agent-spec/pull/38/commits/762bd54b273b5cd27ff1edb010c34626ddc65a6a)

pyagentspec/src/pyagentspec/agent.py
```diff
@@ -54,6 +55,8 @@ class Agent(AgenticComponent):
     """List of toolboxes that are passed to the agent."""
     human_in_the_loop: bool = True
     """Flag that determines if the Agent can request input from the user."""
+    transforms: List[MessageTransform] = Field(default_factory=list)
+    """Additional message transforms that are applied to the messages before they are passed to the agent's LLM."""
```
dhilloulinoracle: can mention the summarization one as an example of message transforms that could be useful when the context becomes long

pyagentspec/src/pyagentspec/transforms/summarization.py
```diff
@@ -0,0 +1,286 @@
...
+    """Maximum number of messages before triggering summarization."""
+
+    min_num_messages: int = 10
+    """Minimum number of recent messages to keep unsummarized."""
```
dhilloulinoracle: can you explain a bit more this parameter and the max one?
I think it would benefit from an example: "if min_num_messages is 10, max_num_messages is 50, then with a conversation of 20 messages, the conversation will not be summarize, but if the conversation has 51 messages, the last 10 won't be summarized, the 41 others will (in one or more summarization messages)"

dhilloulinoracle (approved):

docs/pyagentspec/source/changelog.rst
```diff
@@ -95,6 +95,16 @@ New features
...
 
+* **Datastores**
+
+  Added support for datastores in Agent Spec through :ref:`OracleDatabaseDatastore <oracledatabasedatastore>` and :ref:`PostgresDatabaseDatastore <postgresdatabasedatastore>`.
+  Datastores enable persistent storage and caching capabilities for agent workflows. :ref:`InMemoryCollectionDatastore <inmemorycollectiondatastore>` provides a drop-in replacement for development and prototyping.
+
+* **Context Summarization Transforms**
+
+  Introduced transforms in Agent Spec that allow applying transformations on conversations before being passed to the underlying LLM.
+  We provide :ref:`MessageSummarizationTransform <messagesummarizationtransform>` and :ref:`ConversationSummarizationTransform <conversationsummarizationtransform>` for handling long contexts through summarization.
```
dhilloulinoracle: can we have a how-to?

docs/pyagentspec/source/changelog.rst
```diff
@@ -95,6 +95,16 @@ New features
   Introduced ManagerWorkers in the Agent Spec Language Specification.
   For more information check out the corresponding :doc:`managerworkers how-to guide <howtoguides/howto_managerworkers>` or read the :ref:`API Reference <managerworkers>`.
 
+* **Datastores**
+
+  Added support for datastores in Agent Spec through :ref:`OracleDatabaseDatastore <oracledatabasedatastore>` and :ref:`PostgresDatabaseDatastore <postgresdatabasedatastore>`.
```
dhilloulinoracle: can we have a how-to?

pyagentspec/src/pyagentspec/agent.py
youssef62: Sure, done here [95aecd3](https://github.com/oracle/agent-spec/pull/38/commits/95aecd361ea685d19baeee23cd9a5284be35f535).

pyagentspec/src/pyagentspec/transforms/summarization.py
youssef62: Sure, done here [fdbdc58](https://github.com/oracle/agent-spec/pull/38/commits/fdbdc58a7a65be54206cf09f1bb590162a4730c7). I did not update as I am not sure it's the best place to have examples there, but let me know if you'd like me to.

docs/pyagentspec/source/changelog.rst
youssef62: Sure, I added guides in [aa9098e](https://github.com/oracle/agent-spec/pull/38/commits/aa9098ea989231834d4c75a593cb31d7262fde17). Let me know if you have remaining suggestions.

docs/pyagentspec/source/changelog.rst
youssef62: Sure, I added guides in [aa9098e](https://github.com/oracle/agent-spec/pull/38/commits/aa9098ea989231834d4c75a593cb31d7262fde17). Let me know if you have remaining suggestions.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1946,6 +1955,159 @@ The ManagerWorkers has two main parameters:
...
+        sslcert: Optional[str]
+        sslkey: Optional[SensitiveField[str]]
+        sslrootcert: Optional[str]
+        sslcrl: Optional[str]
+
+- ``user``: User of the postgres database
+- ``password``: Password of the postgres database
+- ``url``: URL to access the postgres database
+- ``sslmode``: SSL mode for the PostgreSQL connection.
+- ``sslcert``: Path of the client SSL certificate, replacing the default `~/.postgresql/postgresql.crt`. Ignored if an SSL connection is not made.
```
paul-cayet:
```suggestion
- ``sslcert``: Path of the client SSL certificate, replacing the default ``~/.postgresql/postgresql.crt``. Ignored if an SSL connection is not made.
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1946,6 +1955,159 @@ The ManagerWorkers has two main parameters:
...
+- ``sslkey``: Path of the file containing the secret key used for the client certificate, replacing the default `~/.postgresql/postgresql.key`. Ignored if an SSL connection is not made.
```
paul-cayet:
```suggestion
- ``sslkey``: Path of the file containing the secret key used for the client certificate, replacing the default ``~/.postgresql/postgresql.key``. Ignored if an SSL connection is not made.
```

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1946,6 +1955,159 @@ The ManagerWorkers has two main parameters:
...
+Summarizes conversations exceeding a given number of messages using an LLM and caches conversation summaries in a ``Datastore``. This is useful to reduce long conversation history into a concise context for downstream LLM calls.
+
+.. code-block:: python
+
+    class ConversationSummarizationTransform(MessageTransform):
+        llm: LlmConfig  # LLM to use for the summarization.
+        max_num_messages: int  # Number of message after which we trigger summarization.
+        min_num_messages: int  # Number of recent messages to keep from summarizing.
+        summarization_instructions: str  # Instruction for the LLM on how to summarize the conversation.
+        summarized_conversation_template: str  # Jinja2 template on how to present the summary (with variable `summary`) to the agent using the transform.
```
paul-cayet:
```suggestion
        summarized_conversation_template: str  # Jinja2 template on how to present the summary (with variable ``summary``) to the agent using the transform.
```

docs/pyagentspec/source/code_examples/howto_datastores.py
```diff
@@ -0,0 +1,149 @@
...
+    "oracle_tls_config_id.dsn": "(DESCRIPTION=(ADDRESS=(PROTOCOL=tcps)(HOST=myhost.example.com)(PORT=2484))(CONNECT_DATA=(SERVICE_NAME=myservice)))",
+    "postgres_config_id.user": "myuser",
+    "postgres_config_id.password": "mypassword",
+    "postgres_config_id.sslkey": "/path/to/client.key",
+}
+
+# Deserialize the configurations
+deserializer = AgentSpecDeserializer()
+
+deserialized_oracle = deserializer.from_yaml(
```
paul-cayet: Please use json instead, same in other places

youssef62: Any reason why to prefer json?

youssef62: Done in [72ef78d](https://github.com/oracle/agent-spec/pull/38/commits/72ef78d46953fc9ad902c6d66ef34a087f2750ce)

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -393,25 +372,20 @@ def _load_field(
 
             # Try to deserialize components/pydantic models according to any of the annotations
             # If any of them works, we will proceed with that. This is our best effort.
+            accumulated_errors = []
             for inner_annotation in inner_annotations:
                 try:
                     return self._load_field(content, inner_annotation)
-                except ValueError:
+                except ValueError as e:
```
paul-cayet, line 379: e not used -> to remove?

PR #39, opened 2025-12-22
Splitting of usage examples in Adapters in different pages

Davide-Cocco:
Divide the Adapter pages into an index (installation guide and brief description) and two pages, one for each usage example (one from AgentSpec to framework, the other is the opposite). 

Deployed docs: https://github.com/oracle/agent-spec/pull/40


nasseur:
Merge Gate INPROGRESS

PR #41, opened 2025-12-23
Feat: tracing for LangGraph flows and add proper async support for LangGraph adapter

cesarebernardis:
Introducing Agent Spec tracing support in the flows of the LangGraph adapter. During conversion, we patch invoke/ainvoke and stream/astream of the CompiledStateGraph to insert the start/end Spans.

This PR also adds support of LangGraph async APIs (e.g. async tools, async LLMs). Async tests are also added for each flow component.

For tracing in the LangGraph adapter, we split the LangGraph event callback into 4 classes: Async/sync x LLM/Tool callbacks. All of them are added during conversion, but the sync ones are inactive if we are under async context (detected via the Trace with a new attribute is_async_mode_active)

Note that this PR might break existing downstream LangGraph tracing span processors if they run in async environment (e.g the AG-UI integration). Span processors will need to implement the async APIs (e.g. `async def on_event_async`) to properly use the async tracing mode.

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
@@ -152,9 +152,8 @@ class AgentSpecCallbackHandler(BaseCallbackHandler):
 
     def _run_in_ctx(self, run_id_str: str, func: Callable[..., T], *args: Any, **kwargs: Any) -> T:
         stack = self._get_stack(run_id_str)
-        current_active_span_stack_token = _ACTIVE_SPAN_STACK.set(stack)
+        _ACTIVE_SPAN_STACK.set(stack)
         result = func(*args, **kwargs)
-        # _ACTIVE_SPAN_STACK.reset(current_active_span_stack_token)
         return result
 
     def _add_event(self, run_id_str: str, span: AgentSpecSpan, event: Any) -> None:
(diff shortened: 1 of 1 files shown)


sonleoracle (approved):

pyagentspec/src/pyagentspec/tracing/trace.py
```diff
@@ -60,6 +60,10 @@ def __init__(
         self.span_processors = span_processors or []
         self.shutdown_on_exit = shutdown_on_exit
         self._root_span = root_span or RootSpan()
+        self._is_async_mode_active: bool = False
```
cesarebernardis, line 63: This is always set, but `is_async_mode_active` seems to be never used, why do you need it?

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -5,9 +5,11 @@
 # (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
 
 import json
+import sys
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+import anyio
```
cesarebernardis, line 12: Is this a new dependency, or do we have it already? If we do an `import` it becomes a direct 3rd party dependency, so we must list it, if it is not already

cesarebernardis: But it does not seem to be used?

pyagentspec/src/pyagentspec/tracing/trace.py
cesarebernardis: My bad, big diff file was collapsed and could not find changes

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: It is already listed in `setup.py` in main, and we are not changing the version in this PR.

it is used for the `anyio.to_thread.run_sync` calls in this file

nasseur:
Internal regression succeeded 🍏: Build ID #261

nasseur:
Internal regression succeeded 🍏: Build ID #269

nasseur:
Internal regression failed: Build ID #359

nasseur:
Merge Gate succeeded 🍏: Build ID #529

PR #43, opened 2026-01-05
feat: MCP tools support in LangGraph adapter

sonleoracle:
# Context

Add support for MCP tools for LangGraph adapter of AgentSpec. Both conversion directions are supported.
Important: an optional dependency was added: `langchain-mcp-adapters`. It is used for converting from AgentSpec to LangGraph as it is the canonical way to create MCP tools in LangChain/LangGraph.

It can be installed with `pip install pyagentspec[langgraph_mcp]`

# Approach

## AgentSpec -> LangGraph

- Added 2 branches (MCPTool and MCPToolBox) in `AgentSpecToLangGraphConverter`
- Core implementation detail: use the async function `load_mcp_tools` from `langchain-mcp-adapters` to discover and load *all* MCP tools from an MCP `connection` (what we call `ClientTransport`). Loaded MCP tools are cached into the `tool_registry` in case the AgentSpecAgent contains separate MCP tools from the same server.
- Because it is async (and our loader is sync), it is wrapped in asyncio.run; if code is executed in a concurrent environment, it is wrapped it in another thread
- The loaded tools are `StructuredTool`s with the `.func` object as a **coroutine**, meaning downstream code must execute the LangGraph graph in an asynchronous setting 
- `langchain-mcp-adapters` supports user-defined `ClientSession`s (via their helper methods), but this PR does not support it yet
- All types of client transports are supported
- Tool filtering in MCP toolboxes are also supported

## LangGraph -> AgentSpec

- Modified `LangGraphToAgentSpecConverter` so that whenever a StructuredTool with a coroutine is encountered, it gets converted to AgentSpecMCPTool.
- It is not possible to convert into a toolbox (since a toolbox are defined by its client transport)
- Core implementation detail: check the `__closure__` of the coroutine to extract the client transport details
- Not all client transports are supported: since the mTLS and HTTPS connections are defined via httpx client factory (it's an arbitrary object), it's very hard to properly convert it to AgentSpec

### DIFF
--- a/docs/pyagentspec/source/changelog.rst
+++ b/docs/pyagentspec/source/changelog.rst
@@ -7,6 +7,12 @@ Agent Spec |release|
 New features
 ^^^^^^^^^^^^
 
+* **MCP tools support in LangGraph adapter:**
+
+  The LangGraph adapter now supports Model Context Protocol (MCP) tools.
+
+  To use this, install the optional extra ``pyagentspec[langgraph_mcp]`` and invoke the loaded graph/agent asynchronously via ``.ainvoke`` within an async context.
+
 * **Added LangGraph adapter to pyagentspec:**
 
   The LangGraph adapter is now available as part of ``pyagentspec``.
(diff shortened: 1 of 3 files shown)


nasseur:
Internal regression succeeded 🍏: Build ID #103

nasseur:
Internal regression succeeded 🍏: Build ID #108

nasseur:
Internal regression succeeded 🍏: Build ID #113

PR #46, opened 2026-01-06
Narrow deserialization exception catching to only value error

paul-cayet:
## Context

When deserializing a Union type field of an Agent Spec component, we try to load the field value by iterating over the possible types, in order. If the serialization fails due to a mismatch (e.g. trying to load a string as a list) we catch the exception and try with the next type.

This PR narrows the try/except block to only catch intentional ValueErrors.

#### Main Changes

- try/except block in deserialization context load field now only catches intentional ValueErrors. 

#### Other changes

- Added types.UnionType to support the pipe syntax for union types
- Fixed incorrect catching when a field value being deserialized is of the wrong type (Property, and when using load_as_dict)

### DIFF
--- a/pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
+++ b/pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
@@ -7,6 +7,7 @@
 """Define the classes and utilities related to deserialization of Agent Spec configurations."""
 
 import inspect
+import types
 import warnings
 from abc import ABC, abstractmethod
 from enum import Enum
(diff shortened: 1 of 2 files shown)


dhilloulinoracle (approved):

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -386,7 +396,7 @@ def _load_field(
             for inner_annotation in inner_annotations:
                 try:
                     return self._load_field(content, inner_annotation)
-                except Exception:  # nosec
+                except ValueError:
```
lfaucon, line 399: @youssef62 potential conflict with your code changes, but it should not be hard to resolve it.

nasseur:
Internal regression succeeded 🍏: Build ID #305

PR #47, opened 2026-01-06
disallow empty titles in properties

soufianeamini:
Add validation to disallow empty titles in properties

P.S: The tests that are removed are not needed anymore since empty titles are caught earlier by this new validation

### DIFF
--- a/docs/pyagentspec/source/changelog.rst
+++ b/docs/pyagentspec/source/changelog.rst
@@ -38,6 +38,13 @@ New features
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
+* **Empty titles in properties**
+
+Property titles in Agent Spec must not be empty. This is now enforced by validation in the pyagentspec SDK.
+
+Migration: If your YAML/JSON configurations have properties without titles, you’ll need to set a non-empty, descriptive title for those properties to pass validation. If you generate Agent Spec configurations via the SDK, your code may still work, but we recommend explicitly setting property titles to ensure forward compatibility.
+
+
 Agent Spec 26.1.0
 -----------------
 
--- a/pyagentspec/src/pyagentspec/property.py
+++ b/pyagentspec/src/pyagentspec/property.py
@@ -89,6 +89,12 @@ class Property(BaseModel):
             self.type = self.json_schema["type"]
 
     @model_validator_with_error_accumulation
+    def _validate_no_empty_titles(self) -> Self:
+        if len(self.title) == 0:
+            raise ValueError(f"Property {self.json_schema} cannot have an empty title")
+        return self
+
+    @model_validator_with_error_accumulation
     def _validate_default(self) -> Self:
         if self.default is Property.empty_default:
             return self
--- a/pyagentspec/tests/test_componentwithio.py
+++ b/pyagentspec/tests/test_componentwithio.py
@@ -46,21 +46,26 @@ def test_component_with_no_inputs_raises_when_input_is_passed() -> None:
         component_cls(name="my_component", inputs=[StringProperty(title="input_1")])
 
 
-def test_component_with_no_inputs_raises_when_empty_input_is_passed() -> None:
-    with pytest.raises(
-        ValueError,
-        match=(
-            r"The MockComponent component received a property titled ``, but did not expect any properties"
-        ),
-    ):
-        component_cls = create_mock_component_cls_with_defaults([], [])
-        component_cls(name="my_component", inputs=[StringProperty(title="")])
-
-
 def test_component_with_inputs_works_when_passed_correct_input() -> None:
     component_cls = create_mock_component_cls_with_defaults([StringProperty(title="input_1")], [])
     component = component_cls(name="my_component", inputs=[StringProperty(title="input_1")])
     assert component.inputs == [StringProperty(title="input_1")]
+
+
+def test_component_with_empty_title_input_raises() -> None:
+    with pytest.raises(ValueError, match=("cannot have an empty title")):
+        component_cls = create_mock_component_cls_with_defaults([StringProperty(title="")], [])
+        component_cls(name="my_component", inputs=[StringProperty(title="")])
+
+
+def test_component_with_empty_input_in_json_schema_raises() -> None:
+    with pytest.raises(ValueError, match=("cannot have an empty title")):
+        component_cls = create_mock_component_cls_with_defaults(
+            [Property(json_schema={"type": "string", "title": ""})], []
+        )
+        component_cls(
+            name="my_component", inputs=[Property(json_schema={"type": "string", "title": ""})]
+        )
 
 
 def test_component_with_inputs_raises_when_missing_inputs_are_not_passed() -> None:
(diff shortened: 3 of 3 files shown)


nasseur:
Internal regression failed: Build ID #106

pyagentspec/tests/test_componentwithio.py
```diff
@@ -46,23 +46,18 @@ def test_component_with_no_inputs_raises_when_input_is_passed() -> None:
...
 def test_component_with_inputs_works_when_passed_correct_input() -> None:
     component_cls = create_mock_component_cls_with_defaults([StringProperty(title="input_1")], [])
     component = component_cls(name="my_component", inputs=[StringProperty(title="input_1")])
     assert component.inputs == [StringProperty(title="input_1")]
 
 
+def test_component_with_empty_input_raises() -> None:
+    with pytest.raises(ValueError, match=("cannot have an empty title")):
+        component_cls = create_mock_component_cls_with_defaults([StringProperty(title="")], [])
+        component_cls(name="my_component", inputs=[StringProperty(title="")])
```
lfaucon, line 58: Nice testing 👍 

Can you also add a check for when the property is defined as `Property(json_schema={'type': 'string', 'title': ''})` ?

lfaucon: I thought we had some issues with the default property title, no? 

```
# (1)
p = Property(json_schema={'type': 'string'})
print(p.title)  # the title equals to "property"
# (2)
q = StringProperty()
print(q.title)  # the title equals to "property"
# (3)
r = Property(json_schema={'type': 'string', 'title': ''}) # This one raises the validation error
# (4)
s = StringProperty(title="") # This one raises the validation error
```

You changed cases (3) and (4) if a manner that is good & well tested.
Do you think that we want to change the behavior of cases (1) and (2) also, or not? It may have a large impact and not be backward compatible :/

soufianeamini: It's an interesting point actually, I think the validation matters more when we want to load a json/yaml configuration that defines the title to be empty, or when the title was explicitly passed as an empty string (which is why the issue was not necessarily with default property title).

Cases (1) and (2) could be kept for convenience when using the SDK to create the configuration from scratch, but we disallow cases (3) and (4) since an (existing) Agent Spec configuration shouldn't be valid if it has properties with empty titles.

If there's a preference for changing those cases as well that's fine by me. We can also make it a warning before it becomes a hard error (regarding backwards compatibility).

pyagentspec/tests/test_componentwithio.py
```diff
@@ -46,47 +46,38 @@ def test_component_with_no_inputs_raises_when_input_is_passed() -> None:
...
 
-def test_component_with_inputs_raises_when_missing_inputs_are_not_passed() -> None:
-    with pytest.raises(
-        ValueError,
-        match=(
-            r"The MockComponent component expected a property titled `input_2`, but none of the passed "
-            r"properties have this title:"
-        ),
-    ):
+def test_component_with_empty_input_raises() -> None:
```
cesarebernardis: Nit: it's not an empty input, it's an input with empty title

pyagentspec/src/pyagentspec/property.py
```diff
@@ -88,6 +88,12 @@ def model_post_init(self, __context: Any) -> None:
         if self.type is None and "type" in self.json_schema:
             self.type = self.json_schema["type"]
 
+    @model_validator_with_error_accumulation
+    def _validate_no_empty_titles(self) -> Self:
```
cesarebernardis, line 92: Did you add the test that @lfaucon asked in https://github.com/oracle/agent-spec/pull/47/files#r2668490145?

I have the feeling that it's going to fail, because here you are checking `self.title`, but in case the title is passed in the json schema, the title attribute is set in the post init, which is supposed to be executed after the validation. If it does not fail, it might be worth to understand how/why, and document it.

pyagentspec/tests/test_componentwithio.py
```diff
@@ -46,47 +46,38 @@ def test_component_with_no_inputs_raises_when_input_is_passed() -> None:
...
+    with pytest.raises(ValueError, match=("cannot have an empty title")):
+        component_cls = create_mock_component_cls_with_defaults([StringProperty(title="")], [])
+        component_cls(name="my_component", inputs=[StringProperty(title="")])
+
+
+def test_component_with_empty_input_in_json_schema_raises() -> None:
+    with pytest.raises(ValueError, match=("cannot have an empty title")):
         component_cls = create_mock_component_cls_with_defaults(
-            [StringProperty(title="input_1"), StringProperty(title="input_2")], []
+            [Property(json_schema={"type": "string", "title": ""})], []
```
cesarebernardis, line 64: Do we have a positive test where we do the same, but with a non-empty title, and we ensure it works correctly?
Asking because of my other comment in the new title validation method. If we do not, let's add one.

pyagentspec/src/pyagentspec/property.py
soufianeamini: Yes I added it in https://github.com/oracle/agent-spec/pull/47/commits/d97214678b886e0d975006068a57adbaf48083db
The test does pass, so I'll look into it to document why.

soufianeamini: The answer is that the model_post_init is actually called before the validation.

pyagentspec/tests/test_componentwithio.py
soufianeamini: Fixed

pyagentspec/tests/test_componentwithio.py
soufianeamini: We have multiple tests where we create Properties the same way but with a non-empty title, but not specifically to see that it doesn't raise errors. But if that behavior changes there are tens of tests that would fail.

pyagentspec/src/pyagentspec/property.py
soufianeamini: From https://github.com/pydantic/pydantic/issues/7163 and https://github.com/pydantic/pydantic/issues/6978#issuecomment-1669907853 we can see that wrap validators are performed after `model_post_init`.

nasseur:
Internal regression succeeded 🍏: Build ID #131

docs/pyagentspec/source/changelog.rst
```diff
@@ -141,7 +141,11 @@ These field will now require to be passed explicitly when loading an exported co
         },
     )
 
+* **Empty titles in properties**
```
dhilloulinoracle: it needs to be in a new 26.2 section;
maybe it was already created, if not please create one

soufianeamini: I've created a new 26.2.0 section 👍

pyagentspec/tests/test_componentwithio.py
soufianeamini: For future reference, we're sticking with a hard error since empty titles were never meant to be empty. We've added it as a breaking change to the changelog of course.

nasseur:
Internal regression failed: Build ID #231

PR #58, opened 2026-01-13
feat: upgrade langgraph to 1.x

sonleoracle:
# Context

We upgrade langgraph and langchain to latest versions since versions < 1.0 are being deprecated.

We also need to migrate the code (esp. the agent spec exporter) as langchain's create_react_agent changed a lot.
MCP tool conversion also need to be migrated as they changed the output signature (it now returns a list of content blocks).

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
@@ -556,7 +556,16 @@ class AgentSpecToolCallbackHandler(AgentSpecCallbackHandler):
         )
         # starting a tool span for this tool
         span_name = f"ToolExecution[{self.tool.name}]"
-        tool_span = AgentSpecToolExecutionSpan(name=span_name, tool=self.tool)
+        # Hack: transmit the tool_call_id as the span's description
+        # so that tool results can learn of its tool_call_id
+        # by correlating with tool starts using the run_id
+        if "tool_call_id" in kwargs:
+            tcid_string = "tcid__" + str(kwargs["tool_call_id"])
+        else:
+            tcid_string = ""
+        tool_span = AgentSpecToolExecutionSpan(
+            name=span_name, description=tcid_string, tool=self.tool
+        )
         self.agentspec_spans_registry[run_id_str] = tool_span
         self._start_and_copy_ctx(run_id_str, tool_span)
         self._add_event(run_id_str, tool_span, request_event)
--- a/pyagentspec/tests/adapters/langgraph/test_tracing.py
+++ b/pyagentspec/tests/adapters/langgraph/test_tracing.py
@@ -12,6 +12,7 @@ from pyagentspec.tracing.events import (
     Event,
     FlowExecutionEnd,
     FlowExecutionStart,
+    LlmGenerationChunkReceived,
     LlmGenerationRequest,
     LlmGenerationResponse,
     NodeExecutionEnd,
@@ -287,3 +288,47 @@ def test_langgraph_stream_tracing_emits_flow_events(json_server: str) -> None:
         assert "a" not in response["outputs"]["haiku_without_a"]
 
     check_dummyspanprocessor_flow_events_and_spans(proc)
+
+
+def test_langgraph_agent_emits_tool_calls_and_results_with_consistent_ids(json_server: str):
+
+    from pyagentspec.adapters.langgraph import AgentSpecLoader
+
+    # Prepare YAML config with placeholders replaced
+    yaml_content = (CONFIGS / "weather_agent_remote_tool.yaml").read_text()
+    final_yaml = _replace_config_placeholders(yaml_content, json_server)
+
+    # Convert to LangGraph agent
+    weather_agent = AgentSpecLoader().load_yaml(final_yaml)
+
+    proc = DummySpanProcessor()
+    with Trace(name="langgraph_tracing_tool_call_test", span_processors=[proc]):
+        agent_input = {
+            "inputs": {},
+            "messages": [{"role": "user", "content": "What's the weather in Agadir?"}],
+        }
+        response = ""
+        for message_chunk, metadata in weather_agent.stream(
+            input=agent_input, stream_mode="messages"
+        ):
+            if message_chunk.content:
+                response += message_chunk.content
+
+    llm_response_chunk_events = [
+        e for (e, _) in proc.events if isinstance(e, LlmGenerationChunkReceived)
+    ]
+    tool_call_chunks = [e for e in llm_response_chunk_events if len(e.tool_calls) == 1]
+    tool_call_ids = {e.tool_calls[0].call_id for e in tool_call_chunks}
+
+    # on_tool_start starts the spans and assigns the tool_call_id to the span's description if available
+    # which is the case in react agents
+    # this allows on_tool_end and ToolExecutionResponse to track the original tool_call_id via the ToolExecutionSpan
+    tool_spans = [s for (_, s) in proc.events if isinstance(s, ToolExecutionSpan)]
+    assert tool_call_ids == set(s.description.replace("tcid__", "") for s in tool_spans)
+
+    # check if the run_ids match
+    tool_start_events = [e for (e, _) in proc.events if isinstance(e, ToolExecutionRequest)]
+    tool_end_events = [e for (e, _) in proc.events if isinstance(e, ToolExecutionResponse)]
+    assert len(tool_start_events) == 1
+    assert len(tool_end_events) == 1
+    assert tool_start_events[0].request_id == tool_end_events[0].request_id


nasseur:
Internal regression succeeded 🍏: Build ID #156

nasseur:
Internal regression succeeded 🍏: Build ID #157

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
sonleoracle, line 230: this PR also upgrades the langchain-mcp-adapters version to latest, which is now fully typed, and we change the code here to address typing issues.

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,76 @@ async def arun():  # type: ignore
...
+            tool_output, list
+        ):  # this is the return type of the tool loaded by langchain_mcp_adapters.tools.load_mcp_tools
+            extracted_values = self._extract_values_from_content_blocks(tool_output)
+            mapped = self._map_extracted_values_to_declared_outputs(extracted_values)
+            return mapped, NodeExecutionDetails()
+        else:
+            # Scalar or other types: map to the single declared output or a generic name
+            return self._map_scalar_to_output(tool_output), NodeExecutionDetails()
+
+    def _extract_values_from_content_blocks(
```
sonleoracle, line 406: since this PR also upgrades langchain-mcp-adapters which changes the output signature of MCP tools, this code (that executes MCP tools) needs to be adapted.

pyagentspec/src/pyagentspec/adapters/langgraph/conversion_utils.py
sonleoracle, line 1: This file is not used anywhere so I removed it.

pyagentspec/setup.py
```diff
@@ -26,12 +26,11 @@ def read(file_name):
 
 
 LANGGRAPH_DEPS = [
-    "langgraph>=0.5.3,<1.0.0",
-    "langchain-core>=0.3,<1.0.0",
-    "langchain-openai>=0.3.7",
-    "langchain-ollama>=0.3.3",
-    "langgraph-checkpoint>=3.0.1,<4.0.0",  # To mitigate CVE-2025-64439
```
cesarebernardis, line 33: Please confirm that langgraph>=1.0.5 requires a version of langgraph-checkpoint newer than 3.0.1, so that we do not need this constraint anymore

pyagentspec/setup.py
```diff
@@ -80,6 +79,6 @@ def read(file_name):
             "autogen-agentchat>=0.5.6; python_version < '3.13'",
         ],
         "langgraph": LANGGRAPH_DEPS,
-        "langgraph_mcp": LANGGRAPH_DEPS + ["langchain-mcp-adapters>=0.1.13,<0.2.0"],
+        "langgraph_mcp": LANGGRAPH_DEPS + ["langchain-mcp-adapters"],
```
cesarebernardis, line 83: What is the version of `langchain-mcp-adapters` for which we have approvals? It should appear in the `constraints.txt` file, but I don't see it.

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,76 @@ async def arun():  # type: ignore
             tool_output = tool(**inputs)
 
         if isinstance(tool_output, dict):
-            # useful for multiple outputs, avoid nesting dictionaries
             return tool_output, NodeExecutionDetails()
+        elif isinstance(
+            tool_output, list
+        ):  # this is the return type of the tool loaded by langchain_mcp_adapters.tools.load_mcp_tools
```
cesarebernardis: How can you be sure that if the output of the tool is a list, then its the output of an mcp tool? What if it's simply a server tool returning a value of type `list[...]`? I think you need to go deeper and check the type/content of the elements.

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,76 @@ async def arun():  # type: ignore
...
+        # No declared outputs
+        return {"tool_output": values}
 
-        output_name = self.node.outputs[0].title if self.node.outputs else "tool_output"
-        return {output_name: tool_output}, NodeExecutionDetails()
+    def _map_scalar_to_output(self, value: Any) -> Dict[str, Any]:
+        node_outputs = self.node.outputs or []
+        if node_outputs:
+            return {node_outputs[0].title: value}
+        return {"tool_output": value}
```
cesarebernardis: I would return an empty dictionary, as reaching this last return would mean the the node does not emit any output

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,76 @@ async def arun():  # type: ignore
...
+                langchain_core_messages_content.ImageContentBlock,
+            ]
+        ],
+    ) -> List[Any]:
+        extracted: List[Any] = []
+        for block in blocks:
+            # Use direct checks on the discriminant to let mypy narrow the TypedDict union
+            if block["type"] == "text":
+                extracted.append(block["text"])
+                continue
```
cesarebernardis: Please use `elif`/`else` and avoid all these `continue`. Or use `case` if you prefer.

pyagentspec/tests/adapters/langgraph/configs/ancestry_agent_with_client_tool.yaml
```diff
@@ -22,10 +22,7 @@ llm_config:
 metadata:
   __metadata_info__: {}
 name: agent_7ced9454
-outputs:
-- default: 0
-  title: result
-  type: integer
+outputs: []
```
cesarebernardis: Why are we removing the tool's output in this test's configuration?

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,76 @@ async def arun():  # type: ignore
...
+            return {node_outputs[0].title: value}
+        if len(node_outputs) > 1:
+            if len(values) != len(node_outputs):
+                raise ValueError(
+                    "MCP tool returned a different number of content blocks than the ToolNode declares as outputs: "
+                    f"returned={len(values)}, declared={len(node_outputs)}"
+                )
+            return {prop.title: values[i] for i, prop in enumerate(node_outputs)}
+        # No declared outputs
+        return {"tool_output": values}
```
cesarebernardis: I would return an empty dictionary, as reaching this last return would mean the the node does not emit any output

pyagentspec/setup.py
sonleoracle: The [bound](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/pyproject.toml) is `"langgraph-checkpoint>=2.1.0,<5.0.0"` so I think we still need the constraint in setup.py

pyagentspec/setup.py
sonleoracle: We should get the approval for the latest version 0.2.1

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
```diff
@@ -317,44 +322,50 @@ def _langgraph_mcp_connection_to_agentspec_client_transport(
                 "Conversion from langchain MCP connections with arbitrary httpx client factory objects is not yet implemented"
             )
 
-        session_params = self._build_session_parameters(conn)
+        session_params = self._build_session_parameters(cast(Mapping[str, Any], conn))
```
soufianeamini: Is the cast still necessary?

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
```diff
@@ -317,44 +322,50 @@ def _langgraph_mcp_connection_to_agentspec_client_transport(
...
 
         if transport == "stdio":
-            c = cast(StdioConnection, conn)
+            c_stdio = cast(StdioConnection, conn)
+            cwd_val = c_stdio.get("cwd")
+            cwd_str: Optional[str]
+            if isinstance(cwd_val, Path):
+                cwd_str = str(cwd_val)
+            else:
+                cwd_str = cwd_val
```
soufianeamini: I think you can simplify these lines to the following:
```suggestion
            c_stdio = cast(StdioConnection, conn)
            cwd_str = str(conn.get("cwd"))
```

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
```diff
@@ -317,44 +322,50 @@ def _langgraph_mcp_connection_to_agentspec_client_transport(
...
 
         if transport == "streamable_http":
-            c = cast(StreamableHttpConnection, conn)
+            c_http = cast(StreamableHttpConnection, conn)
             return StreamableHTTPTransport(
                 name="agentspec_streamablehttp_transport",
-                url=c["url"],
-                headers=c.get("headers"),
+                url=c_http["url"],
+                headers=cast(Optional[Dict[str, str]], c_http.get("headers")),
```
soufianeamini: I think you can remove the casts here

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,146 @@ async def arun():  # type: ignore
...
+            return {}
+
+        # 1 declared output
+        if len(node_outputs) == 1:
+            out_name = node_outputs[0].title
+
+            if output_type == "dict":
+                # If raw_output is exactly {out_name: <value>}, keep as-is; otherwise wrap the whole dict
+                if isinstance(raw_output, dict) and out_name in raw_output and len(raw_output) == 1:
+                    return {out_name: raw_output[out_name]}
```
soufianeamini: Considering you check that `len(raw_output) == 1`, you then don't need to recreate the dict since it only contains that element

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,146 @@ async def arun():  # type: ignore
...
+                return {out_name: raw_output}
+
+            if output_type in list_types:
+                # Use first value if exactly one, else the whole list/tuple
+                value = (
+                    raw_output[0]
+                    if hasattr(raw_output, "__len__") and len(raw_output) == 1
+                    else raw_output
+                )
+                return {out_name: value}
```
soufianeamini: How does this work if the property is a list that just so happens to have a single element in some cases? Other nodes might still perform list operations on the value

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -283,11 +284,146 @@ async def arun():  # type: ignore
...
+            return {out_name: raw_output}
+
+        # Multiple declared outputs
+        if output_type == "dict":
+            declared_titles = {prop.title for prop in node_outputs}
+            return {
+                k: raw_output[k]
+                for k in declared_titles
+                if isinstance(raw_output, dict) and k in raw_output
+            }
```
soufianeamini: The isinstance check can be called once outside the dict comprehension. Would also make it explicity that we return an empty dict if `raw_output` is not a dict (assuming this is the intent, if not then we might want to raise an explicit error for missing outputs)

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: updated the tool output mapping logic to make it clearer

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: updated the tool output mapping logic to make it clearer

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: updated the tool output mapping logic to make it clearer

pyagentspec/tests/adapters/langgraph/configs/ancestry_agent_with_client_tool.yaml
sonleoracle: Reverted the change. This is the agent's output (not the client tool's output). I originally changed it because with the new langgraph update, the behavior of the agent with structured outputs (Agent.outputs is specified) is that it would try to make a tool call and a tool message to return the correct output type. I will change the client tool test instead to capture this new behavior.

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: updated the tool output mapping logic to make it clearer

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
sonleoracle: Nope, removed it

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
sonleoracle: changed

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: narrowed down so that this is only applied when it's a list with 1 element returned by the MCP tool, and when the output type is not list

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: this isinstance check is not neccessary because it is already done when this helper is called

pyagentspec/src/pyagentspec/adapters/langgraph/_agentspecconverter.py
sonleoracle: removed

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
sonleoracle: yes I return the `raw_output` directly

nasseur:
Internal regression succeeded 🍏: Build ID #179

nasseur:
Merge Gate INPROGRESS

PR #73, opened 2026-01-16
update version file

aadnaane:
update version file in the release branch


lfaucon (approved):
Should we also in this PR:

(1) Update https://github.com/oracle/agent-spec/blob/main/pyagentspec/src/pyagentspec/versioning.py#L33 to mention 26.1.1 as the new current version?

(2) Update the changelog, such that is has a title saying "Agent Spec 26.1.0" with the content of the version `26.1.0` and the changelog should be reset to empty below the title "Agent Spec |release|"  https://github.com/oracle/agent-spec/blob/main/docs/pyagentspec/source/changelog.rst so that it gets filled in the next weeks with the new features for the next release

nasseur:
Internal regression succeeded 🍏: Build ID #184

ElayachiFouad (approved):

PR #77, opened 2026-01-20
Draft: feat: Add Tool Confirmation for LangGraph adapter

vikhyatt:
**Context**: Tool Confirmation has been present in PyAgentSpec, but was not supported for the LangGraph adapter. 

**Changes**: This PR tries a best-effort support for LangGraph adapter by passing Tools (or nodes) requiring confirmation in the `interrupts_before` flag such that the execution is interrupted before the Tool is called. The user can then choose to stop the Tool from executing if needed.


sonleoracle:
This feature was merged in this PR #85

PR #79, opened 2026-01-20
feat: [LangGraph adapter] add support for disaggregated configs

### LINKED ISSUE #78: Support disaggregated components in the LangGraph adapter
# Context

Disaggregated configs is a new security feature that recently was introduced to avoid including [sensitive information](https://oracle.github.io/agent-spec/development/agentspec/language_spec_nightly.html#sensitive-fields) in Agent Spec configurations. As such, when serializing Agent Spec components with sensitive fields (e.g., API keys), these fields must be provided in the `components_registry` when deserializing it back to the component.

The WayFlow adapter supports this feature, but not the LangGraph adapter. Let's add support for this feature for the LangGraph adapter.

# Acceptance

Users can use the `pyagentspec.adapters.langgraph.AgentSpecLoader` to load an Agent Spec component with disaggregated attributes:

```python
AgentSpecLoader(tool_registry={...}).load_json(agent_json, components_registry={"llm_config.api_key": "<MY API KEY>"})
```

sonleoracle:
Context: Agent Spec recently introduced disaggregated configs during serialization/deserialization to address sensitive fields (e.g. LLM API keys). We should support that in the LangGraph adapter.

closes #78

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
@@ -9,11 +9,14 @@ from typing import Any, Dict, List, Optional, Union, cast, overload
 
 from pyagentspec.adapters.langgraph._langgraphconverter import AgentSpecToLangGraphConverter
 from pyagentspec.adapters.langgraph._types import (
+    BaseChatModel,
     Checkpointer,
     CompiledStateGraph,
     LangGraphComponentsRegistryT,
     LangGraphRuntimeComponent,
     RunnableConfig,
+    StateGraph,
+    StructuredTool,
 )
 from pyagentspec.component import Component as AgentSpecComponent
 from pyagentspec.serialization import AgentSpecDeserializer, ComponentDeserializationPlugin
@@ -312,11 +315,9 @@ class AgentSpecLoader:
 
         converted: Dict[str, Any] = {}
         for custom_id, value in registry.items():
-            if isinstance(value, AgentSpecComponent):
-                converted[custom_id] = value
-            else:
-                # this converts the value into a langgraph object
-                # if conversion fails (e.g. unsupported langgraph object), it raises a NotImplementedError
+            if isinstance(value, (StateGraph, CompiledStateGraph, BaseChatModel, StructuredTool)):
                 converted[custom_id] = converter.convert(value)
+            else:
+                converted[custom_id] = value
 
         return converted


docs/pyagentspec/source/changelog.rst
```diff
@@ -113,6 +113,14 @@ Improvements
 
   Introduced support for Python version 3.14.
 
+* **Disaggregated configurations in LangGraph adapter**
```
cesarebernardis: Release is already out, this should end up in a new section. Please:
- Copy the `Agent Spec |release|` section with all its subsections and remove the content
- Make this one `Agent Spec 26.1.0`
- Copy this new changelog entry in the new section you copied

nasseur:
Internal regression succeeded 🍏: Build ID #198

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -28,33 +29,137 @@ def __init__(
...
+        components_registry: Optional[AgentSpecComponentsRegistryT],
+    ) -> CompiledStateGraph[Any, Any, Any]: ...
+
+    @overload
+    def load_yaml(
+        self,
+        serialized_assistant: str,
+        *,
+        import_only_referenced_components: bool,
+    ) -> Union[CompiledStateGraph[Any, Any, Any], Dict[str, AgentSpecComponent]]: ...
```
cesarebernardis: This looks a bit weird: the main component is converted, but the referenced components are not? That's inconsistent.

If you look at what `wayflow` does at https://github.com/oracle/wayflow/blob/main/wayflowcore/src/wayflowcore/agentspec/runtimeloader.py#L137, everything gets transformed. It has a drawback though, like https://github.com/oracle/wayflow/blob/main/wayflowcore/src/wayflowcore/agentspec/runtimeloader.py#L248 , which is that it requires to convert back when you have a `components_registry`, as you expect that it contains LangGraph components instead of Agent Spec ones

nasseur:
Internal regression succeeded 🍏: Build ID #206

docs/pyagentspec/source/changelog.rst
sonleoracle: Done

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
sonleoracle: I updated the code to make it closely follow wayflow, by converting `components_registry` in a best-effort manner. Please take another look, thanks.

docs/pyagentspec/source/changelog.rst
```diff
@@ -4,6 +4,20 @@ Changelog
...
+
+  The LangGraph adapter now supports loading disaggregated components during deserialization.
+  Use `AgentSpecLoader.load_yaml/json(..., import_only_referenced_components=True)` to import
+  referenced components (e.g., LLM configs, tools), optionally modify them, and then pass them back
+  via `components_registry` when loading the main configuration. This enables keeping sensitive
+  fields (like API keys) out of the main spec while still resolving them at load time.
+
+Agent Spec 26.1.0
+--------------------
```
cesarebernardis, line 19: Nit: length of the `---` stripe should match the one of the title

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -28,54 +55,276 @@ def __init__(
...
+        >>> from pyagentspec.llms import OllamaConfig
+        >>> from pyagentspec.serialization import AgentSpecSerializer
+        >>> agent = Agent(id="agent_id", name="A", system_prompt="You are helpful.", llm_config=OllamaConfig(name="m", model_id="llama3.1", url="http://localhost:11434"))
+        >>> main_json, disag_json = AgentSpecSerializer().to_json(
+        ...     agent, disaggregated_components=[(agent.llm_config, "llm_id")], export_disaggregated_components=True
+        ... )
+        >>> from pyagentspec.adapters.langgraph import AgentSpecLoader
+        >>> loader = AgentSpecLoader()
+        >>> registry = loader.load_json(disag_json, import_only_referenced_components=True)
+        >>> compiled = loader.load_json(main_json, components_registry=registry)
```
cesarebernardis: To avoid double conversion of the elements in the registry, maybe in these examples we could show that we deserialize the disaggregated components with the pyagentspec deserializer and we pass that to the `components_registry`

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -28,54 +55,276 @@ def __init__(
...
+
+        converted: Dict[str, Any] = {}
+        for custom_id, value in registry.items():
+            # If it's already an Agent Spec component/value, or not a known LangGraph runtime type,
+            # pass through unchanged so users can mix registries.
+            if isinstance(value, AgentSpecComponent) or not (
+                isinstance(value, (StateGraph, CompiledStateGraph, BaseChatModel, StructuredTool))
+            ):
+                converted[custom_id] = value
+                continue
```
cesarebernardis: Please avoid using `continue`

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -28,54 +55,276 @@ def __init__(
...
+
+            # Prefer centralized converter for supported runtime values
+            converted_value = converter.convert_runtime_value(value)
```
cesarebernardis: I don't get:
- the value of this separate new method, why don't we just call convert
- why we are using the LangGraphToAgentSpecConverter and the AgentSpecExporter separately, aren't they supposed to do the same thing (just with a few abstraction layers of difference)

Can you elaborate?

sonleoracle: Because the `AgentSpecExporter._convert` method assumes the top-level object is already a graph (while `value` here can be a langgraph llm, structured tool, etc. So I think either we use a helper method, like here, to convert each langgraph type separately (with routing to type-specific conversion methods) or we refactor `AgentSpecExporter._convert` so that its recursion base cases are better.

cesarebernardis: You are adding the `convert_runtime_value` to the `LangGraphToAgentSpecConverter`, not to the `AgentSpecExporter`, and `LangGraphToAgentSpecConverter` already has `convert` that does not suffer of the problem you are mentioning.

`AgentSpecExporter.to_component` is just calling `LangGraphToAgentSpecConverter.convert` adding a check on the object to be converted (this is what you are talking about), but I don't see a reason to use it here. If you don't have either, please just use `LangGraphToAgentSpecConverter.convert` and remove the method `LangGraphToAgentSpecConverter.convert_runtime_value`.

docs/pyagentspec/source/changelog.rst
sonleoracle: fixed

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
sonleoracle: good idea, I added an example like that

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
sonleoracle: changed

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
sonleoracle: I merged this helper into `._convert` and now I use `LangGraphToAgentSpecConverter.convert` to convert the registry components.

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -28,54 +52,271 @@ def __init__(
...
-            ),
+        from pyagentspec.adapters.langgraph._agentspecconverter import (
+            LangGraphToAgentSpecConverter,
         )
+
+        converter = LangGraphToAgentSpecConverter()
+
+        converted: Dict[str, Any] = {}
+        for custom_id, value in registry.items():
+            if isinstance(value, AgentSpecComponent):
```
cesarebernardis: I know that it's harder, but please do the opposite and keep the original value as part of the `else`. In disaggregated components we might have values that are not agent spec components (e.g., dictionaries, strings, etc.) and they do not need to be converted.

sonleoracle: Done

nasseur:
Internal regression succeeded 🍏: Build ID #337

nasseur:
Merge Gate succeeded 🍏: Build ID #503

PR #86, opened 2026-01-26
Wayflow adapter in pyagentspec

cesarebernardis:
Expose `wayflowcore` 's `AgentSpecLoader` and `AgentSpecExporter` in `pyagentspec`, so that Wayflow can be used like the other agentic framework adapters simply through `pyagentspec`

### DIFF
--- a/pyagentspec/tests/test_schema_generation.py
+++ b/pyagentspec/tests/test_schema_generation.py
@@ -154,7 +154,7 @@ def test_generated_schema_correctly_validates_all_valid_serializations(
     file_path: str, component_type: Type[Component]
 ) -> None:
     serialized_component = read_agentspec_config_file(file_path)
-    component_schema = component_type.model_json_schema()
+    component_schema = component_type.model_json_schema(only_core_components=True)
     validate(yaml.safe_load(serialized_component), component_schema)
 
 


docs/pyagentspec/source/installation.rst
```diff
@@ -70,7 +70,8 @@ Extra dependencies
 |project| offers optional extra dependencies that can be installed to enable additional features.
 
 * The ``autogen`` extra dependency gives access to the AutoGen runtime adapter.
-* The ``langgraph`` extra dependency gives access to the LangGraph runtime adapter.
+* The ``langgraph``, ``langgraph_mcp`` extra dependencies give access to the LangGraph runtime adapter.
+* The ``wayflow`` extra dependency gives access to the WayFlow runtime adapter.
```
jschweiz: nit: you can maybe cite the optional dependencies like the langgraph row above (`wayflow[oci]` for example)

pyagentspec/setup.py
```diff
@@ -81,5 +81,6 @@ def read(file_name):
         ],
         "langgraph": LANGGRAPH_DEPS,
         "langgraph_mcp": LANGGRAPH_DEPS + ["langchain-mcp-adapters>=0.1.13,<0.2.0"],
+        "wayflow": ["wayflowcore>=25.4.3; python_version < '3.14'"],
```
jschweiz, line 84: is that the lower bound `25.4.3`? I don't think 25.4.3 will work due to mismatches in the adapters with this version of pyagentspec

pyagentspec/src/pyagentspec/adapters/wayflow/agentspecexporter.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing import TYPE_CHECKING
+
+from pyagentspec._lazy_loader import LazyLoader
+
+if TYPE_CHECKING:
+    # Important: do not move this import out of the TYPE_CHECKING block so long as wayflow is an optional dependency.
+    # Otherwise, importing the module when they are not installed would lead to an import error.
+
+    from wayflowcore.agentspec import AgentSpecExporter
```
sonleoracle, line 15: So pyagentspec.adapters.wayflow is just a convenience for the actual agentspec module of wayflowcore?

pyagentspec/setup.py
sonleoracle, line 84: What about the oci optional dependency of wayflowcore?

cesarebernardis: As we don't have a common interface across loaders (yet) this should not be a problem. I agree that we should use 26.1, but it's not out yet. We'll update as soon as it gets released.

pyagentspec/src/pyagentspec/adapters/wayflow/agentspecexporter.py
cesarebernardis: Yes

docs/pyagentspec/source/installation.rst
cesarebernardis: I'll add the extra deps that are required for the components supported in agent spec (i.e., all of them `oci,a2a,datastores`)

pyagentspec/setup.py
cesarebernardis: Added

dhilloulinoracle (approved):

PR #93, opened 2026-02-06
Agent Spec Evaluation

cesarebernardis:
Introducing Open Agent Specification Evaluation, a new extension of Agent Spec that standardizes how agentic systems are evaluated in a framework-agnostic way.

### DIFF
--- a/pyagentspec/src/pyagentspec/evaluation/metrics/llm_based_metric.py
+++ b/pyagentspec/src/pyagentspec/evaluation/metrics/llm_based_metric.py
@@ -47,6 +47,10 @@ class LlmBasedMetric(Metric[MetricValueType]):
     async def ask_llm(self, conversation: List[Dict[str, str]]) -> Tuple[str, Tuple[int, int]]:
         """Return the assistant message text and token usage from the LLM provider."""
         response = await self._complete_conversation(conversation)
+        if "choices" not in response or len(response["choices"]) == 0:
+            raise RuntimeError(
+                f"LLM returned an empty response during the computation of {self.name}"
+            )
         return (
             response["choices"][0]["message"]["content"],
             (response["usage"]["prompt_tokens"], response["usage"]["completion_tokens"]),


docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+It defines a common semantic and minimal APIs for:
+
+- Datasets: collections of samples to evaluate against.
+- Metrics: reusable, composable measurements (LLM-based or deterministic) with retry and exception handling.
+- Evaluators: orchestrators that apply metrics to datasets with concurrency control.
+
+Agent Spec Eval emphasizes
+- Separation of concerns (dataset vs. metrics vs. orchestration);
+- Evaluation transparency (details, timing, tokens, failed attempts);
+- Robustness to LLM non-determinism (retries, repetition, ensembles, exception handling);
```
dhilloulinoracle: is the "ensembles" part referring to "ensembles of judges"? Might be good to say it in this case, would be clearer

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+Repeat: recompute a metric multiple times and aggregate. Useful for noisy LLM-based metrics.
+
+.. code-block:: python
+
+   class RepeatMetric(Metric[AggregationType]):
+     metric: Metric[RepeatMetric]
+     aggregator: Aggregator[MetricValueType, AggregationType]
+     num_repeats: int
+
+Ensemble: run semantically equivalent metrics and aggregate. Useful to sue different LLMs or prompts for the same metric.
```
dhilloulinoracle: > Useful to sue different LLMs

I guess there is a typo here; don't sue me if I am wrong

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+.. code-block:: python
+
+   class EnsembleMetric(Metric[AggregationType]):
+     metrics: Sequence[Metric[T]]
+     aggregator: Aggregator[MetricValueType, AggregationType]
+
+Intermediates
+-------------
+
+Some metrics depend on shared intermediate values (e.g., a list of facts extracted from a response).
```
dhilloulinoracle: I don't think they really "depend" on shared values. Maybe we can say "Some metrics need similar intermediate values (e.g. list of facts extracted from a response) in which cases using 'intermediates' enables computing those once and re-use them for computing the different metrics that need them"

docs/pyagentspec/source/agentspec/evaluation.rst
dhilloulinoracle, line 1: this seems to explain in details the capabilities, but maybe is a bit rough for people trying to determine how to setup some evaluation.
Can we have a how-to?

pyagentspec/setup.py
```diff
@@ -89,5 +89,11 @@ def read(file_name):
             "agent-framework==1.0.0b251028; python_version < '3.14'",
             "agent-framework-core==1.0.0b251028; python_version < '3.14'",
         ],
+        "evaluation": [
+            "anyio>=4.10.0,<4.12.0",
+            "litellm>=1.79.0",
```
sonleoracle: Do we have approvals for `litellm`?

pyagentspec/tests/conftest.py
```diff
@@ -103,6 +103,11 @@ def default_llm_config() -> VllmConfig:
     )
 
 
+@pytest.fixture(scope="session")
```
sonleoracle, line 122: This one is already added in conftest of langgraph tests, please remove it from there if needed

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+- Parallelism controls to respect rate limits;
+- Portability across runtimes and agent frameworks.
+
+Where appropriate, Agent Spec Eval can interoperate with Agent Spec Tracing to collect traces of LLM/tool activity during evaluation
+(see :ref:`agentspec_tracing`).
+
+The objectives of Agent Spec Eval are:
+- Provide a canonical set of evaluation components and their interfaces.
+- Define lifecycle and attribute schemas for metrics, datasets, and evaluators.
+- Identify sensitive fields and how they should be handled (aligning with Agent Spec security guidance).
```
sonleoracle: What does this mean?

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Core Concepts
+=============
+
+Sample
+------
+
+A Sample is a dictionary-like structure containing features used by metrics.
+Samples should be identified by an ``id``, and they can contain arbitrary data.
+Common keys include:
```
sonleoracle, line 42: Common keys? What are some other special keys?

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+
+- ``query``: the user request or task.
+- ``response``: the model/agent output to be evaluated.
+- ``reference``: a gold/ground-truth reference, when available.
+- ``conversation`` (optional): for multi-turn evaluations, the full interaction serialized, e.g., as a list of messages in OpenAI format
```
sonleoracle, line 47: Is it mandatory to use the OpenAI format here? (I think yes)

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+       Orchestrates retries, timing, and exception handling around compute_metric.
+       Applies input_mapping (if provided) to map sample feature names to parameter names.
+       Returns (value_or_fallback, details).
+       """
+
+Return value semantics:
+
+- If the Metric call returns a value, it is considered a valid measurement (no retry triggered).
+- If an attempt fails, raise an exception; the wrapper will retry up to ``num_retries``.
+- If all attempts fail, ``on_failure`` decides the fallback:
```
sonleoracle, line 118: What's the default fallback?

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+
+   class Evaluator:
+     metrics: Sequence[Metric[Any]]
+     max_concurrency: int = -1
+
+     async def evaluate(self, dataset: Dataset) -> EvaluationResults: ...
+
+- ``max_concurrency`` controls how many metric computations run at once.
+  - ``-1`` means "auto" (no explicit limiter by the Evaluator).
+  - Use a positive integer to avoid rate-limit errors with LLM backends.
```
sonleoracle:
```suggestion
  - Use a positive integer to avoid rate-limit errors with LLM providers when using LLM-as-a-Judge metrics.
```

docs/pyagentspec/source/agentspec/evaluation.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Design Notes:
+
+- When ``max_concurrency`` is small, throughput is bounded but safer (fewer rate-limit errors).
+- When ``-1``, concurrency is unconstrained by the Evaluator (relying on backend libraries or runtime limits).
+
+Examples
+========
+
+End-to-end evaluation
```
sonleoracle, line 230: Please manually test these examples to make sure they work

docs/pyagentspec/source/installation.rst
```diff
@@ -73,6 +73,7 @@ Extra dependencies
 * The ``langgraph``, ``langgraph_mcp`` extra dependencies give access to the LangGraph runtime adapter.
```
sonleoracle: Unrelated to this PR but could you change it to 

```suggestion
* The ``langgraph``, ``langgraph-full`` extra dependencies give access to the LangGraph runtime adapter.
```

examples/evaluation/10_ensemble_metric.py
```diff
@@ -0,0 +1,69 @@
...
+    ("Zeurich", "Zurich"),
+    ("Beijing", "Peking"),
+    ("Knight", "Night"),
+    ("Write", "Right"),
+]
+
+
+async def main() -> None:
+    llms = {
+        "llama_3": "oci/meta.llama-3.3-70b-instruct",
```
sonleoracle: I guess this `oci/` prefix is for litellm. It is not clear we are using litellm under the hood, and that it requires the `model_id` to have the provider as the prefix. Can we document this somewhere?

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
```diff
@@ -0,0 +1,111 @@
...
+            pass
+        else:
+            raise TypeError(f"Unknown `features_consistency` mode: {features_consistency}.")
+
+        if not features:
+            raise ValueError("No features found under the selected consistency policy.")
+
+        return features
+
+    async def get_sample(self, id: Any) -> Dict[str, Any]:
```
sonleoracle: Why does this have to be async?

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
```diff
@@ -0,0 +1,111 @@
...
+        return self.data[id]
+
+    def features(self) -> Collection[str]:
+        return self._features
+
+    async def ids(self) -> AsyncIterator[Any]:
```
sonleoracle: Why does this have to be async?

examples/evaluation/11_aggregator.py
```diff
@@ -0,0 +1,35 @@
...
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+from typing import Collection
+
+from pyagentspec.evaluation.aggregators import Aggregator
+
+
+class MyMeanAboveThresholdAggregator(Aggregator[float | int | bool, bool]):
```
sonleoracle: This is quite complex type hinting, can we try to make it simpler for users to write custom subclasses?

examples/evaluation/3_dataset_special_key.py
```diff
@@ -0,0 +1,26 @@
...
+    "Q1": {"query": "Where is the largest city of CH?", "ref": "Zürich", "res": "Zurich"},
+    "Q2": {"query": "Where is the capital of Switzerland?", "ref": "Bern", "res": "Bern"},
+    "Q3": {"query": "Where is the UN European headquarter?", "ref": "Geneva", "res": "Genève"},
+}
+dataset = Dataset.from_dict(data)
+
+
+async def main() -> None:
+    async for id in dataset.ids():
+        sample = await dataset.get_sample(id)
```
sonleoracle, line 21: See comment below discussing why this has to be async

examples/evaluation/5_llm_call.py
```diff
@@ -0,0 +1,52 @@
...
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+import asyncio
+import os
+
+from litellm.llms.custom_httpx.async_client_cleanup import close_litellm_async_clients
+
+from pyagentspec.evaluation._llm.invocation import complete_conversation
```
sonleoracle: why are we importing from a private module in a demo script?

pyagentspec/setup.py
```diff
@@ -89,5 +89,11 @@ def read(file_name):
...
+            "pandas>=2.0.3,<3.0.0",
+            "oci>=2.158.2",
```
sonleoracle, line 135: so `oci` is a third-party dependency now? do we have approvals for it?

pyagentspec/src/pyagentspec/evaluation/intermediates/intermediate.py
```diff
@@ -0,0 +1,60 @@
...
+    @abstractmethod
+    async def compute_value(
+        self, *args: Any, **kwargs: Any
+    ) -> Tuple[IntermediateValueType, Dict[str, Any]]:
+        """Compute the intermediate value and return it with metadata details."""
+        raise NotImplementedError(
+            "Method `compute_value` must be implemented for any subclass of `Intermediate`."
+        )
+
+    async def __call__(self, **kwargs: Any) -> Tuple[IntermediateValueType, Dict[str, Any]]:
```
sonleoracle: This `__call__` definition does not allow `*args` here, but in `compute_metric` of the metric classes, it supports `*args` and `**kwargs`? Would it be a good idea if we force `**kwargs` everywhere?

pyagentspec/src/pyagentspec/evaluation/datasets/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+        -------
+        int
+            Number of samples.
+        """
+
+        return len(self._data_source)
+
+    @staticmethod
+    def from_dict(
+        data: Dict[Any, Dict[str, Any]] | List[Dict[str, Any]],
```
sonleoracle: Should be a Hashable type instead of Any

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluation_results.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Container for storing and accessing evaluation metric results for multiple samples and metrics.
+
+    This class provides utilities to work with evaluation results that are organized as a mapping between
+    (sample_id, metric_name) pairs and their corresponding result values and details. It enables exporting
+    the results to common formats such as JSON and pandas DataFrame for further analysis or reporting.
+
+    Attributes
+    ----------
+    results : Dict[Tuple[str, str], Tuple[Any, Dict[str, Any]]]
```
sonleoracle: Isn't the sample_id supposed to be any hashable type instead of a `str` (e.g. see Dataset.from_dict)

pyagentspec/src/pyagentspec/evaluation/datasets/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+        Any
+            Unique identifier for a sample.
+        """
+
+        return self._data_source.ids()
+
+    def __len__(self) -> int:
+        """
+        Synchronously get the number of available samples.
+        Usually wraps `alen()`.
```
sonleoracle: Please either remove this line or

```suggestion
        Wraps `len()`.
```

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+
+    def _process_attempts_result(
+        self,
+        failed_attempts: Sequence[EvaluationException],
+        successful_attempt: Tuple[MetricValueType, Dict[str, Any]] | None,
+        computation_details: Dict[str, Any] | None = None,
+    ) -> Tuple[MetricValueType | None, Dict[str, Any]]:
+        injecting_details = {
+            "__failed_attempts": failed_attempts,
```
sonleoracle, line 148: If we are dumping results to json, I think we will get an error when trying to serialize the `EvaluationException` exception objects. Please add a test where the results of a failed evaluation run are saved to json.

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluation_results.py
```diff
@@ -0,0 +1,105 @@
...
+        metric_names : List[str], optional
+            List of metric names. If not provided, inferred from the metric names present in the results dictionary.
+
+        """
+
+        self.results = results
+        self.sample_ids = sample_ids or list({sample_id for sample_id, _ in self.results.keys()})
+        self.metric_names = metric_names or list({m_name for _, m_name in self.results.keys()})
+
+    def to_json(self) -> Dict[str, Dict[str, Dict[str, Any]]]:
```
sonleoracle: `to_json` implies it returns a string, whereas here it returns a dict. I think this method should be called `to_dict`.

pyagentspec/src/pyagentspec/evaluation/aggregators/implementations/mean_aggregator.py
```diff
@@ -0,0 +1,42 @@
...
+        float
+            Arithmetic mean computed as ``sum(values) / len(values)``.
+
+        Notes
+        -----
+        Users should not invoke :meth:`aggregate` directly. Call the aggregator instance itself,
+        e.g. ``aggregator(values)``.
+        """
+        _values = [float(v) for v in values]
+        return sum(_values) / len(_values)
```
sonleoracle, line 46: what if `_values` is an empty list? this would raise `ZeroDivisionError`, which might be confusing to users

pyagentspec/src/pyagentspec/evaluation/aggregators/implementations/harmonic_mean_aggregator.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Notes
+        -----
+        Users should not invoke :meth:`aggregate` directly. Call the instance itself instead
+        (e.g. ``aggregator(values)``).
+        """
+        _values = [float(v) for v in values]
+
+        if any(v < 0 for v in _values):
+            raise RuntimeError(
```
sonleoracle: Is there a test for this check?

pyagentspec/src/pyagentspec/evaluation/metrics/llm_based_metric.py
```diff
@@ -0,0 +1,53 @@
...
+
+    async def _complete_conversation(self, conversation: List[Dict[str, str]]) -> Dict[str, Any]:
+        """Send ``conversation`` to the configured LLM and return the raw payload."""
+        return await complete_conversation(conversation, self.llm_config)
+
+    async def ask_llm(self, conversation: List[Dict[str, str]]) -> Tuple[str, Tuple[int, int]]:
+        """Return the assistant message text and token usage from the LLM provider."""
+        response = await self._complete_conversation(conversation)
+        return (
+            response["choices"][0]["message"]["content"],
```
sonleoracle, line 55: This key-based indexing assumes the `response` object is a dict with the OpenAI API style payload. Can we use base models instead (maybe import base models from `litellm` or define our own)? Otherwise, can we add checks here?

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/with_intermediate_metric.py
```diff
@@ -0,0 +1,53 @@
...
+
+from typing import Any, Dict, Literal, Sequence, Tuple
+
+from pyagentspec.evaluation._utils import _get_duplicates
+from pyagentspec.evaluation.exceptions.handling_strategies import ExceptionHandlingStrategy
+from pyagentspec.evaluation.intermediates import Intermediate
+from pyagentspec.evaluation.metrics.metrics import Metric, MetricValueType
+
+
+class WithIntermediatesMetric(Metric[MetricValueType | None]):
```
sonleoracle, line 15: This public class has no docstring, please add a docstring for it

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/ensemble_metric.py
```diff
@@ -0,0 +1,97 @@
...
+    ) -> None:
+        """
+        Initialize the EnsembleMetric.
+
+        Parameters
+        ----------
+        name : str
+            The name for this ensemble metric.
+
+        metrics : Collection[Metric[T]]
```
sonleoracle: type hint in docstring does not match type hint in method signature

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/with_intermediate_metric.py
```diff
@@ -0,0 +1,53 @@
...
+            on_failure=on_failure,
+        )
+        self.intermediates = intermediates
+        self.metric = metric
+
+    async def compute_metric(
+        self, *args: Any, **kwargs: Any
+    ) -> Tuple[MetricValueType | None, Dict[str, Any]]:
+        intermediates_values = {
+            intermediate.name: (await intermediate(*args, **kwargs))[0]
```
sonleoracle, line 62: See [this comment](https://github.com/oracle/agent-spec/pull/93/changes#r2811677427)

`Intermediate.__call__` does not allow `*args` but the call here passes `*args`? I think it would crash.

pyagentspec/src/pyagentspec/evaluation/intermediates/computing.py
```diff
@@ -0,0 +1,54 @@
...
+    computer = _AsyncCallablesComputer(
+        dataset=dataset,
+        callables={intermediate.name: intermediate for intermediate in intermediates},
+        max_concurrency=max_concurrency,
+    )
+    results = await computer.run()
+    return Dataset.from_dict(
+        data={
+            sample_id: {
+                intermediate.name: _result_to_dict(results[(sample_id, intermediate.name)])
```
sonleoracle: `_result_to_dict` stores each intermediate as `{ "value": ..., "details": ... }`, so a dataset enhanced via `add_intermediates()` ends up with entries like `{intermediate.name: {"value": ..., "details": ... }}`. When the same intermediate is injected inline using `WithIntermediatesMetric`, the metric receives `{intermediate.name: <raw value>}` (check the code). Please unify these representations (either make both paths return the raw value and stash details separately, or have the wrapper pass the same `{value, details}` dict) so consuming metrics don’t have to special-case depending on how intermediates were produced.

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
```diff
@@ -0,0 +1,94 @@
...
+                fingerprint=config_file["fingerprint"],
+                private_key_file_location=config_file["key_file"],
+                pass_phrase=config_file.get("pass_phrase"),
+            ),
+        }
+
+    raise NotImplementedError(f"OciClientConfig type not supported: {type(client_config)}")
+
+
+def _get_llm_config_as_dict(llm: LlmConfig) -> Dict[str, Any]:
```
sonleoracle: Since this is kind of a `litellm` adapter for Agent Spec, please rename this method to `_llm_convert_to_litellm`

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: Rephrased

pyagentspec/tests/conftest.py
cesarebernardis: It was already removed from there

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: Added

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: No, the docs say it's an example, it does not force this structure

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
cesarebernardis: Fixed and added a test

pyagentspec/src/pyagentspec/evaluation/datasets/dataset.py
cesarebernardis: I think that the intent of the docstring is correct, but I will remove to avoid confusion

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: Unrelated, I will remove the line

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: This is not a strict definition, we leave the format open.

docs/pyagentspec/source/agentspec/evaluation.rst
cesarebernardis: Specified

examples/evaluation/10_ensemble_metric.py
cesarebernardis: Good point, I don't think we should ask users to put `oci`, it should be Agent Spec - like. I changed the code to don't need this `oci` prefix

examples/evaluation/11_aggregator.py
cesarebernardis: This is a class defined in an example, people should not extend this

cesarebernardis: I simplified it a bit anyway

examples/evaluation/5_llm_call.py
cesarebernardis: Yeah, probably this example does not make much sense. I'll remove it.

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
cesarebernardis: Renamed to include litellm

pyagentspec/src/pyagentspec/evaluation/aggregators/implementations/mean_aggregator.py
cesarebernardis: Made it raise a specific exception in case of empty list

pyagentspec/src/pyagentspec/evaluation/datasets/dataset.py
cesarebernardis: Fixed

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
cesarebernardis: This was by design. I think that the main reason is to be able to run evaluation in an async manner (e.g., to avoid sync locks on remote llm calls).

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
cesarebernardis: This was by design. I think that the main reason is to be able to run evaluation in an async manner (e.g., to avoid sync locks on remote llm calls).

pyagentspec/src/pyagentspec/evaluation/aggregators/implementations/harmonic_mean_aggregator.py
cesarebernardis: Added

examples/evaluation/3_dataset_special_key.py
cesarebernardis: This was by design. I think that the main reason is to be able to run evaluation in an async manner (e.g., to avoid sync locks on remote llm calls).

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
sonleoracle: Yeah I get it, but should these specific methods be async too (`self.ids`, `self. get_sample`)? Can they be sync?

examples/evaluation/3_dataset_special_key.py
sonleoracle: See comment https://github.com/oracle/agent-spec/pull/93#discussion_r2812870866

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
cesarebernardis: In this case yes, but since it's the implementation of the basic interface, that is `async` as you don't necessarily know how heavy it is to retrieve a sample. I would keep it as is, since it's probably a minor detail.

examples/evaluation/3_dataset_special_key.py
cesarebernardis: Answered

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
```diff
@@ -0,0 +1,216 @@
...
+        bound_args = _bind_kwargs_to_func(self.compute_metric, *args, **kwargs)
+
+        failed_attempts: List[EvaluationException] = []
+        for attempt_id in range(1 + self.num_retries):
+            try:
+                time_attempt_start = time.time()
+                val, val_details = await self.compute_metric(*bound_args.args, **bound_args.kwargs)
+                time_attempt_end = time.time()
+
+                logger.log(
```
paul-cayet: Supernit but that's exactly logger.info(...)

pyagentspec/src/pyagentspec/evaluation/intermediates/computing.py
```diff
@@ -0,0 +1,54 @@
...
+                for intermediate in intermediates
+            }
+            async for sample_id in dataset.ids()
+        }
+    )
+
+
+async def add_intermediates(
```
paul-cayet, line 36: Is that a public function? Can we have a description and a short code example in the docstring of this function?

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
```diff
@@ -0,0 +1,216 @@
...
+
+        """
+
+        _on_failure: ExceptionHandlingStrategy | None
+        if on_failure == "raise":
+            _on_failure = Raise()
+        elif on_failure == "set_none":
+            _on_failure = SetConstant(None)
+        elif on_failure == "set_zero":
+            _on_failure = SetConstant(0)
```
paul-cayet, line 87: Apart from the type casting output types are not enforced/validated, do we want to add this type validation? (or later if there is a need to?)

pyagentspec/src/pyagentspec/evaluation/_computers/_async_callables_computers.py
```diff
@@ -0,0 +1,125 @@
...
+
+    async def run(self) -> Dict[Tuple[Any, str], T]:
+        """Kick off all pending computations and return the populated registry."""
+        # Materialise identifiers up-front to avoid holding async generators open
+        # while scheduling the computation fan-out.
+        sample_ids = [sample_id async for sample_id in self.dataset.ids()]
+        metrics_names = list(self.callables.keys())
+        # ``anyio`` drives every (sample, metric) pair while respecting
+        # the concurrency limit enforced by ``_queue``.
+        async with anyio.create_task_group() as tg:
```
paul-cayet, line 129: The concurrency limit is on the _compute, but here you can still create a massive amounts of tasks, can we improve this?

pyagentspec/src/pyagentspec/evaluation/_computers/_async_callables_computers.py
```diff
@@ -0,0 +1,125 @@
...
+            for sample_id in sample_ids:
+                for metric_name in metrics_names:
```
paul-cayet: Are we ok with failure in 1 task cancelling the whole group? Should we do some kind of exception collection and "late raising"?

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluator.py
```diff
@@ -0,0 +1,101 @@
...
+        self,
+        metrics: Sequence[Metric[Any]],
+        max_concurrency: int = -1,
+    ) -> None:
+        """
+        Initializes the Evaluator with a collection of metrics and concurrency settings.
+
+        Parameters
+        ----------
+        metrics : Sequence[Metric[Any]])
```
paul-cayet:
```suggestion
        metrics : Sequence[Metric[Any]]
```

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluator.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        computer = _AsyncCallablesComputer[Tuple[Any, Dict[str, Any]]](
+            dataset=dataset,
+            callables={metric.name: metric for metric in self.metrics},
+            max_concurrency=self.max_concurrency,
+        )
+        results = await computer.run()
+        return EvaluationResults(
+            results=results,
+            sample_ids=[id_ async for id_ in dataset.ids()],
```
paul-cayet, line 99: FYI computer.run builds sample_ids already (might not be a problem but so that you know)

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluator.py
```diff
@@ -0,0 +1,101 @@
...
+        Returns
+        -------
+        EvaluationResults
+            Structured view over the metric values and their associated metadata.
+
+        Notes
+        -----
+        Metrics run concurrently whenever ``max_concurrency`` permits. Any
+        :class:`pyagentspec.evaluation.exceptions.EvaluationException` raised by an underlying
+        metric propagates to the caller.
```
paul-cayet: Nit: This is true only if on_failure=Raise() right? otherwise the exceptions are caught?

pyagentspec/src/pyagentspec/evaluation/aggregators/implementations/harmonic_mean_aggregator.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class HarmonicMeanAggregator(Aggregator[bool | float | int, float]):
+    """
+    Aggregator that computes the harmonic mean of a collection of non-negative numerical values.
+
+    Call an instance of this class with a sequence of non-negative numbers (bool, int, or float) to obtain their harmonic mean.
+
+    If any value is zero, the result is zero.
+    Negative values will raise a RuntimeError.
```
paul-cayet:
```suggestion
    Negative values will raise a ValueError.
```

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
```diff
@@ -0,0 +1,111 @@
...
+        else:
+            raise TypeError(f"Unknown `features_consistency` mode: {features_consistency}.")
+
+        if not features:
+            raise ValueError("No features found under the selected consistency policy.")
+
+        return features
+
+    async def get_sample(self, id: Any) -> Dict[str, Any]:
+        return self.data[id]
```
paul-cayet: Nit: can have slightly nicer error than the base KeyError?

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
```diff
@@ -0,0 +1,94 @@
...
+    if isinstance(llm, OpenAiConfig):
+        return {"model": llm.model_id}
+    if isinstance(llm, OpenAiCompatibleConfig):
+        url = llm.url
+        if not url.startswith("http"):
+            url = f"http://{url}"
+        if not url.endswith("v1"):
+            url += "/v1"
+        return {"model": "openai/" + llm.model_id, "api_base": url}
+    raise NotImplementedError(f"LlmConfig type not supported: {type(llm)}")
```
paul-cayet: Nit
```suggestion
    raise NotImplementedError(
        f"LlmConfig type not supported: {type(llm)}. Supported types are "
        "OciGenAiConfig, OpenAiConfig, and OpenAiCompatibleConfig.
    )
```

pyagentspec/src/pyagentspec/evaluation/metrics/llm_as_a_judge_metric.py
```diff
@@ -0,0 +1,183 @@
...
+
+from pyagentspec.adapters._utils import render_template
+from pyagentspec.evaluation.exceptions import EvaluationException
+from pyagentspec.evaluation.exceptions.handling_strategies import ExceptionHandlingStrategy
+from pyagentspec.evaluation.metrics.llm_based_metric import LlmBasedMetric
+from pyagentspec.evaluation.metrics.metrics import MetricValueType
+from pyagentspec.llms import LlmConfig
+from pyagentspec.templating import get_placeholders_from_string
+
+logger = logging.getLogger(__name__)
```
paul-cayet: nit: unused

pyagentspec/src/pyagentspec/evaluation/intermediates/intermediate.py
```diff
@@ -0,0 +1,60 @@
...
+        name: str,
+        input_mapping: Dict[str, str] | None = None,
+    ) -> None:
+        """
+        Store the intermediate name and optional dataset field mapping.
+
+        Parameters
+        ----------
+        name
+            The name of the metric.
```
paul-cayet:
```suggestion
            The name of the intermediate.
```

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/with_intermediate_metric.py
cesarebernardis: Added

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/ensemble_metric.py
cesarebernardis: Fixed

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
cesarebernardis: Replaced

pyagentspec/src/pyagentspec/evaluation/intermediates/computing.py
cesarebernardis: Added

pyagentspec/src/pyagentspec/evaluation/datasets/_dict_data_source.py
cesarebernardis: Done

pyagentspec/src/pyagentspec/evaluation/metrics/llm_as_a_judge_metric.py
cesarebernardis: Removed

pyagentspec/setup.py
cesarebernardis: In progress, they will be needed for the release, not a blocker for now

pyagentspec/setup.py
cesarebernardis: In progress, they will be needed for the release, not a blocker for now

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluation_results.py
cesarebernardis: Agree, updated

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluation_results.py
cesarebernardis: Updated

pyagentspec/src/pyagentspec/evaluation/_computers/_async_callables_computers.py
cesarebernardis: Improved

pyagentspec/src/pyagentspec/evaluation/intermediates/intermediate.py
cesarebernardis: I think that the intent was to have `*args` everywhere, and it was forgotten here. I'll add it here.

pyagentspec/src/pyagentspec/evaluation/metrics/wrappers/with_intermediate_metric.py
cesarebernardis: Answered in the other comment

pyagentspec/src/pyagentspec/evaluation/metrics/metrics.py
cesarebernardis: This PR is already quite big, I would keep it for later

pyagentspec/src/pyagentspec/evaluation/_computers/_async_callables_computers.py
cesarebernardis: I think that supporting something like this would require to answer a few other questions and solve problems. Just collecting the exceptions and raise at the end would not help much, as the caller would not have access anyway to the results (not easily at least), a failure would happen in the end, and it would take a long time to notice that. Moreover, (expected) failures should already be covered by the metric itself, so if we have to raise here, it means that we are actually supposed to raise, or there was another major issue. I would keep it as-is for now, we might review this later if needed.

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluator.py
cesarebernardis: Changed, now it does not anymore

pyagentspec/src/pyagentspec/evaluation/evaluator/evaluator.py
cesarebernardis: Yes, updated

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
```diff
@@ -0,0 +1,97 @@
...
+            "model": "oci/" + llm.model_id,
+            "oci_compartment_id": llm.compartment_id,
+            "oci_serving_mode": llm.serving_mode,
+            **_get_oci_client_config(llm.client_config),
+        }
+    if isinstance(llm, OpenAiConfig):
+        return {"model": llm.model_id}
+    if isinstance(llm, OpenAiCompatibleConfig):
+        url = llm.url
+        if not url.startswith("http"):
```
sonleoracle: I often ran into trouble when giving model endpoints to litellm, so we should make sure it is robust here

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+    raise NotImplementedError(f"OciClientConfig type not supported: {type(client_config)}")
+
+
+def _get_llm_config_as_litellm_dict(llm: LlmConfig) -> Dict[str, Any]:
+    """Convert an ``LlmConfig`` instance into provider-specific invocation kwargs."""
+    if isinstance(llm, OciGenAiConfig):
+        return {
+            "model": "oci/" + llm.model_id,
```
sonleoracle, line 76: what if llm.model_id already starts with `oci/` ?

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
```diff
@@ -0,0 +1,97 @@
...
+        }
+    if isinstance(llm, OpenAiConfig):
+        return {"model": llm.model_id}
+    if isinstance(llm, OpenAiCompatibleConfig):
+        url = llm.url
+        if not url.startswith("http"):
+            url = f"http://{url}"
+        if not url.endswith("v1"):
+            url += "/v1"
+        return {"model": "openai/" + llm.model_id, "api_base": url}
```
sonleoracle: what if llm.model_id already starts with `openai/`?

sonleoracle (approved):

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
cesarebernardis: It's a wrong model id then, it should follow agent spec rules

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
cesarebernardis: It's a wrong model id, it should follow agent spec rules

pyagentspec/src/pyagentspec/evaluation/_llm/invocation.py
cesarebernardis: You should not give endpoints to litellm, you should follow agent spec rules. Users should not care about what we use to make llm calls

pyagentspec/src/pyagentspec/evaluation/intermediates/computing.py
cesarebernardis: Done, uniformed to just the value

docs/pyagentspec/source/agentspec/evaluation.rst
dhilloulinoracle: can we put those code examples in python files that get run in the CI?

cesarebernardis: Done

nasseur:
Internal regression failed: Build ID #358

nasseur:
Merge Gate failed: Build ID #599

pyagentspec/src/pyagentspec/evaluation/metrics/llm_based_metric.py
cesarebernardis: It's the `to_dict` of a `ModelResponse` returned by a `acompletion` call of litellm, it should be ok as-is. I'll just add a couple of checks.

nasseur:
Internal regression succeeded 🍏: Build ID #417

Issue #94, opened 2026-02-07
feat: Anthropic Support for Langgraph


f4roukb:
**Description**
AgentSpec doesn't support converting from/to Langgraph when the underlying LLM inference client is `ChatAnthropic`. For now, it only explicitly supports OLLaMa, OpenAI, and OpenAI-compatible LLM inference clients.

**Feature Request**
Please introduce support for Anthropic LLM inference client similarly to the other AI providers/inference frameworks. I think this would be a strong addition to drive adoption since Anthropic has [>30% market share in B2B](https://dataconomy.com/2025/08/04/anthropic-leads-enterprise-llms-with-32-market-share/?utm_source=chatgpt.com).

dhilloulinoracle:
Hey,
Anthropic models are supported through the new Generic LLM support: https://github.com/oracle/agent-spec/issues/104 .
So will close this ticket

PR #95, opened 2026-02-07
feat: Support Anthropic for Langgraph

f4roukb:
**Issue:**  https://github.com/oracle/agent-spec/issues/94

**Description**: Introduce basic support for Anthropic LLM inference client. Happy to extend this in the future as-needed.

**Changes**:
- **New dependency:** Add the `langchain-anthropic` dependency to the `langgraph` extra.
- Add a new LLM config.
- Add unit tests for `ChatAnthropic` conversion between AgentSpec and Langgraph.
- Update docs.
**No breaking changes.**


pyagentspec/src/pyagentspec/llms/anthropicconfig.py
```diff
@@ -0,0 +1,25 @@
+# Copyright © 2025 Oracle and/or its affiliates.
```
dhilloulinoracle, line 1: probably needs to be `2026` here

dhilloulinoracle (approved):

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
```diff
@@ -0,0 +1,25 @@
...
+"""Defines the class for configuring how to connect to Anthropic Claude models."""
+
+from pyagentspec.llms.llmconfig import LlmConfig
+
+
+class AnthropicLlmConfig(LlmConfig):
+    """
+    Class to configure a connection to an Anthropic Claude model.
+
+    Requires to specify the model identity. The API key and endpoint are optional
```
cesarebernardis: Since you mention API keys here, could you add to this class an attribute marked as sensitive for it?
You can copy the `api_key` you find in the `OpenAiConfig`

cesarebernardis (approved):

pyagentspec/tests/adapters/langgraph/llms/test_anthropic_llm_conversion.py
sonleoracle, line 1: Please rename this new test folder to `pyagentspec/tests/adapters/langgraph/llms`

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
```diff
@@ -0,0 +1,25 @@
...
+    and may be provided by the runtime environment.
+    """
+
+    model_id: str
+    """ID of the model to use."""
+
+    base_url: str | None = None
```
sonleoracle: When are users expected to provide a non-default base URL? Are Anthropic models hosted on different endpoints? And would those different versions be compatible with the `ChatAnthropic` from `LangChain`? The existing `OpenAiConfig` does not allow users to specify an URL.

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
```diff
@@ -0,0 +1,25 @@
...
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+"""Defines the class for configuring how to connect to Anthropic Claude models."""
+
+from pyagentspec.llms.llmconfig import LlmConfig
+
+
+class AnthropicLlmConfig(LlmConfig):
```
sonleoracle, line 17: You could add the `api_key` sensitive field here as well in the `AnthropicLlmConfig`, following the `OpenAiConfig`. However, the LangGraph adapter does not yet convert directly this `api_key` field into the langchain chat models; instead it still expects the api key environment variable to be defined **before** the conversion.

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Class to configure a connection to an Anthropic Claude model.
+
+    Requires to specify the model identity. The API key and endpoint are optional
+    and may be provided by the runtime environment.
+    """
+
+    model_id: str
+    """ID of the model to use."""
```
sonleoracle:
```suggestion
    """ID of the Anthropic model to use, e.g., claude-haiku-4-5-20251001"""
```

sonleoracle (changes requested):

sonleoracle (approved):
Thank you for your contribution @f4roukb !

f4roukb:
Thanks both! I'll have a look at the OCA as soon as possible and get back to you.

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
f4roukb: Thanks Cesare.

Please correct me if I'm wrong but the conversion from Langgraph to AgentSpec for the supported LLM clients drops the API key. The created LLM config won't have the API key unless set post conversion.

If I were to add an api_key param, I would have to add a check to set it only if available when converting from AgentSpec to Langgraph to avoid preventing the client from picking up the key from the env. This introduces hidden behavior that could be error-prone.

My goal was to avoid having to deal with two potential sources of truth for what the API key can be so I didn't introduce the parameter.

What do you think?

Edit: As-is right now, I'll remove the mention of the API key.

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
f4roukb: Thanks Son. Let's align here in the conversation with Cesare.

https://github.com/oracle/agent-spec/pull/95#pullrequestreview-3772439305

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
f4roukb: Exposing the base URL would be something you might need to support for integration/component tests for products based on top of your framework.

If the base_url isn't exposed, this doesn't prevent but unnecessarily complexifies writing and maintaining such tests.

Let me know what you think!

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
cesarebernardis: I understand the concern, but this is the definition at Agent Spec level, and it requires to think one level higher. 

> Please correct me if I'm wrong but the conversion from Langgraph to AgentSpec for the supported LLM clients drops the API key. The created LLM config won't have the API key unless set post conversion.

That's correct for the state being, and that's an acceptable solution, but the LangGraph -> Agent Spec conversion is one way to create Agent Spec configurations. Moreover, the api_key would be removed in case the configuration gets serialized, since it is a sensitive field, so I would not worry about:

> This introduces hidden behavior that could be error-prone.

The main concern is not this direction of the conversion, but the opposite, i.e., from Agent Spec to LangGraph (or any framework). Ideally, the api key can be provided as part of the components registry at deserialization time, and this api key should be used to create the LLM object. This aligns with what most of the agentic frameworks allow in their LLM classes.

pyagentspec/src/pyagentspec/llms/anthropicconfig.py
cesarebernardis: > The existing OpenAiConfig does not allow users to specify an URL.

That's true, but for the same reason we have the `OpenAiCompatibleConfig`.
I think we can go one of these two ways:
- Keep the url in here (and I would call it `url` to align with the other models)
- Create an `AnthropicCompatibleConfig` that has a url

sonleoracle: I would go with keeping it as `url`

cesarebernardis (approved):
LGTM, thank you for the contribution!

cesarebernardis:
Don't worry about the failures in the tests. As soon as you will have signed the OCA, I will take care of finalizing and merging the PR.

cesarebernardis:
Supported through Generic LLM configuration

dhilloulinoracle:
Hey @f4roukb ,
We recently merged https://github.com/oracle/agent-spec/pull/147 that enables support for all types of LLMs in a generic manner.
That should fill the purpose of this PR.

Please have a look, and let us know what you think.

PR #100, opened 2026-02-13
Typescript SDK

### LINKED ISSUE #97: TypeScript SDK for Agent Spec
The only SDK available today for the Agent Spec is in Python. It would be great to have a TypeScript SDK so the JS/TS ecosystem can also author, validate, and serialize Agent Spec configurations.                                                                                                                       

The initial phase would cover the core spec (components, properties, serialization, validation), with runtime adapters following in later phases.

I'm happy to contribute this if the team is open to it.

spichen:
Resolves #97

### DIFF
--- a/pyagentspec/setup.py
+++ b/pyagentspec/setup.py
@@ -34,6 +34,9 @@ LANGGRAPH_DEPS = [
     "anyio>=4.10.0,<4.12.0",
     "langgraph-swarm>=0.1.0",
     # 4rth party dependencies
+    # langgraph-prebuilt 1.0.9 imports runtime symbols that are not present in
+    # the langgraph 1.0.x series we currently support.
+    "langgraph-prebuilt<1.0.9",
     "langgraph-checkpoint>=3.0.1,<4.0.0",  # To mitigate CVE-2025-64439
 ]
 
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/_types.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/_types.py
@@ -4,13 +4,11 @@
 # (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
 # (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
 
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Callable, Dict, Hashable, Tuple, TypedDict, Union
 
 from typing_extensions import TypeAlias
 
 from pyagentspec._lazy_loader import LazyLoader
-from pyagentspec.llms.openaicompatibleconfig import OpenAIAPIType
 
 if TYPE_CHECKING:
     # Important: do not move this import out of the TYPE_CHECKING block so long as langgraph is an optional dependency.
@@ -20,36 +18,27 @@ if TYPE_CHECKING:
     import langchain_core.messages.content as langchain_core_messages_content
     import langchain_ollama
     import langchain_openai
-    import langgraph
     import langgraph.graph as langgraph_graph
-    import langgraph.graph.state as langgraph_graph_state
-    import langgraph.prebuilt as langgraph_prebuilt
-    import langgraph.types as langgraph_types
     import langgraph_swarm
     from langchain.agents.middleware.types import AgentState
     from langchain_core.callbacks import BaseCallbackHandler
     from langchain_core.language_models import BaseChatModel
     from langchain_core.messages import BaseMessage, SystemMessage, ToolMessage
     from langchain_core.outputs import ChatGenerationChunk, GenerationChunk, LLMResult
-    from langchain_core.runnables import RunnableBinding, RunnableConfig, RunnableLambda
+    from langchain_core.runnables import RunnableConfig, RunnableLambda
     from langchain_core.tools import BaseTool, StructuredTool
     from langgraph.graph import StateGraph
     from langgraph.graph._branch import BranchSpec
     from langgraph.graph._node import StateNodeSpec
     from langgraph.graph.message import Messages
     from langgraph.graph.state import CompiledStateGraph
-    from langgraph.store.base import BaseStore
     from langgraph.types import Checkpointer, interrupt
 
 else:
-    langgraph = LazyLoader("langgraph")
     langgraph_swarm = LazyLoader("langgraph_swarm")
     langchain_ollama = LazyLoader("langchain_ollama")
     langchain_openai = LazyLoader("langchain_openai")
     langgraph_graph = LazyLoader("langgraph.graph")
-    langgraph_types = LazyLoader("langgraph.types")
-    langgraph_prebuilt = LazyLoader("langgraph.prebuilt")
-    langgraph_graph_state = LazyLoader("langgraph.graph.state")
     langchain_agents = LazyLoader("langchain.agents")
     langchain_core_messages_content = LazyLoader("langchain_core.messages.content")
     # We need to import the classes this way because it's the only one accepted by the lazy loader
@@ -61,18 +50,14 @@ else:
     Messages = LazyLoader("langgraph.graph.message").Messages
     CompiledStateGraph = LazyLoader("langgraph.graph.state").CompiledStateGraph
     BaseCallbackHandler = LazyLoader("langchain_core.callbacks").BaseCallbackHandler
-    AsyncCallbackHandler = LazyLoader("langchain_core.callbacks").AsyncCallbackHandler
-    RunnableBinding = LazyLoader("langchain_core.runnables").RunnableBinding
     RunnableConfig = LazyLoader("langchain_core.runnables").RunnableConfig
     RunnableLambda = LazyLoader("langchain_core.runnables").RunnableLambda
     StateNodeSpec = LazyLoader("langgraph.graph._node").StateNodeSpec
-    StateNode = LazyLoader("langgraph.graph._node").StateNode
     BranchSpec = LazyLoader("langgraph.graph._branch").BranchSpec
     SystemMessage = LazyLoader("langchain_core.messages").SystemMessage
     BaseMessage = LazyLoader("langchain_core.messages").BaseMessage
     ToolMessage = LazyLoader("langchain_core.messages").ToolMessage
     BaseChatModel = LazyLoader("langchain_core.language_models").BaseChatModel
-    BaseStore = LazyLoader("langgraph.store.base").BaseStore
     ChatGenerationChunk = LazyLoader("langchain_core.outputs").ChatGenerationChunk
     GenerationChunk = LazyLoader("langchain_core.outputs").GenerationChunk
     LLMResult = LazyLoader("langchain_core.outputs").LLMResult
@@ -83,14 +68,6 @@ LangGraphTool: TypeAlias = Union[BaseTool, Callable[..., Any]]
 LangGraphComponent = Union[StateGraph[Any, Any, Any], CompiledStateGraph[Any, Any, Any]]
 LangGraphRuntimeComponent: TypeAlias = Union[LangGraphComponent, BaseChatModel, StructuredTool]
 LangGraphComponentsRegistryT: TypeAlias = Dict[str, Union[LangGraphRuntimeComponent, Any]]
-
-
-@dataclass
-class LangGraphLlmConfig:
-    model_type: str
-    model_name: str
-    base_url: str
-    api_type: OpenAIAPIType = OpenAIAPIType.CHAT_COMPLETIONS
 
 
 class NodeExecutionDetails(TypedDict, total=False):
@@ -129,12 +106,8 @@ ControlFlow: TypeAlias = Dict[SourceNodeId, Dict[BranchName, TargetNodeId]]
 
 
 __all__ = [
-    "langgraph",
     "langgraph_graph",
-    "langgraph_graph_state",
-    "langgraph_types",
     "langchain_core_messages_content",
-    "langgraph_prebuilt",
     "langchain_agents",
     "langchain_ollama",
     "langchain_openai",
@@ -142,7 +115,6 @@ __all__ = [
     "LangGraphComponent",
     "LangGraphRuntimeComponent",
     "LangGraphComponentsRegistryT",
-    "LangGraphLlmConfig",
     "NodeExecutionDetails",
     "FlowStateSchema",
     "FlowInputSchema",
@@ -157,7 +129,6 @@ __all__ = [
     "CompiledStateGraph",
     "StateGraph",
     "StateNodeSpec",
-    "RunnableBinding",
     "StructuredTool",
     "BaseTool",
     "RunnableLambda",
@@ -165,7 +136,6 @@ __all__ = [
     "BaseMessage",
     "ToolMessage",
     "BaseChatModel",
-    "BaseStore",
     "AgentState",
     "Checkpointer",
     "interrupt",
@@ -173,7 +143,6 @@ __all__ = [
     "Messages",
     "BranchSpec",
     "BaseCallbackHandler",
-    "AsyncCallbackHandler",
     "ChatGenerationChunk",
     "GenerationChunk",
     "LLMResult",


dhilloulinoracle:
Thanks a lot @spichen for this big contribution!
We will start reviewing it.

To be able to merge it, we would need you to sign the CLA. Would it be possible for you to have a look at that?

tsagentspec/src/flows/edges/data-flow-edge.ts
```diff
@@ -0,0 +1,35 @@
...
+  componentType: z.literal("DataFlowEdge"),
+  sourceNode: NodeRef,
+  sourceOutput: z.string(),
+  destinationNode: NodeRef,
+  destinationInput: z.string(),
+});
+
+export type DataFlowEdge = z.infer<typeof DataFlowEdgeSchema>;
+
+export function createDataFlowEdge(opts: {
```
sonleoracle, line 26: `createDataFlowEdge` never checks that `sourceOutput` exists on `sourceNode` or that `destinationInput` exists on `destinationNode`. The Python SDK enforces this with Pydantic validators, preventing misspelled property connections. Can you add equivalent checks (and tests) to keep flow graphs valid at construction time?

tsagentspec/tests/property.test.ts
sonleoracle, line 1: The suite covers scalar properties comprehensively but does not exercise `objectProperty` with nested schemas or `dictProperty` with complex value types. A couple of extra assertions would confirm that `validateSchemaTitle` recurses correctly.

tsagentspec/src/flows/nodes/llm-node.ts
```diff
@@ -0,0 +1,53 @@
...
+  description?: string;
+  metadata?: Record<string, unknown>;
+  inputs?: Property[];
+  outputs?: Property[];
+}): LlmNode {
+  const inputs =
+    opts.inputs ??
+    getPlaceholderPropertiesFromJsonObject(opts.promptTemplate);
+
+  const outputs = opts.outputs ?? [
```
sonleoracle, line 34: Default output property is hand-constructed as a wrapper `{ jsonSchema, title, type, ... }` rather than using the property constructors. This increases the chance of drift (e.g., missing title validation, missing schema normalization). Consider reusing a `stringProperty(...)` constructor.

tsagentspec/src/component.ts
```diff
@@ -0,0 +1,113 @@
...
+import { z } from "zod";
+import { PropertySchema } from "./property.js";
+
+/** Base component schema - all components extend this */
+export const ComponentBaseSchema = z.object({
+  id: z.string().uuid().default(() => crypto.randomUUID()),
+  name: z.string(),
+  description: z.string().optional(),
+  metadata: z.record(z.unknown()).default({}),
+  componentType: z.string(),
```
sonleoracle, line 16: Internal field name is `componentType`, but serialization uses `component_type`. That’s workable, but it increases drift risk and already causes bugs in deserialization (nested component detection doesn’t respect `camelCase`). Consider aligning internal field to the canonical serialized discriminator or ensuring a single, consistent mapping layer.

tsagentspec/src/flows/nodes/llm-node.ts
spichen: Done

tsagentspec/tests/property.test.ts
spichen: Done

tsagentspec/src/flows/edges/data-flow-edge.ts
spichen: Done

tsagentspec/src/property.ts
```diff
@@ -0,0 +1,435 @@
...
+    )) {
+      if (typeof inner === "object" && inner !== null) {
+        validateSchemaTitle(inner as JsonSchemaValue);
+      }
+    }
+  }
+}
+
+/** Base property Zod schema */
+export const PropertySchema = z.object({
```
sonleoracle, line 64: `PropertySchema` defaults `jsonSchema` to `{}` and `title` to `"property"`, and `type` is optional. This allows constructing a `Property` that serializes to JSON Schema missing required spec annotations (`title`, `type`). Recommend enforcing `jsonSchema.title` and `jsonSchema.type` invariants at construction time.

tsagentspec/src/serialization/deserialization-context.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (loaded instanceof DeserializationInProgressMarker) {
+      throw new Error(
+        `Circular dependency during deserialization of object with id: '${referenceId}'`,
+      );
+    }
+    return loaded!;
+  }
+
+  /** Load a single field value, handling component refs and nested structures */
+  loadField(value: unknown): unknown {
```
sonleoracle, line 113: `loadField()` only treats nested component dicts as components when `"component_type" in dict`. If `camelCase` mode is enabled (or user-provided nested dicts use `componentType`), nested components will be treated as plain objects and won’t deserialize properly. Suggest using the same key selection logic as `getComponentType()` (`camelCase ? "componentType" : "component_type"`).

tsagentspec/src/serialization/serialization-context.ts
```diff
@@ -0,0 +1,294 @@
...
+    return result;
+  }
+
+  /** Serialize a component to a dict, handling referencing */
+  dumpComponentToDict(component: ComponentBase): ComponentAsDict {
+    const componentId = component.id;
+    const mappedId =
+      this.componentsIdMapping.get(componentId) ?? componentId;
+
+    if (!this.resolvedComponents.has(mappedId)) {
```
sonleoracle, line 190: Potential cache key inconsistency for disaggregation: checks `!this.resolvedComponents.has(mappedId)` but later stores `this.resolvedComponents.set(componentId, componentDump)`. If `mappedId !== componentId` (disaggregated or remapped IDs), lookups can fail and `$component_ref` may point at an ID not present in `$referenced_components`.

tsagentspec/src/serialization/serialization-context.ts
```diff
@@ -0,0 +1,294 @@
...
+ * Not a general-purpose converter — uppercase segments after underscores are not matched.
+ */
+export function snakeToCamel(str: string): string {
+  return str.replace(/_([a-z0-9])/g, (_, letter: string) =>
+    letter.toUpperCase(),
+  );
+}
+
+/** Fields that should never be transformed between camelCase/snake_case */
+const NEVER_TRANSFORM_FIELDS = new Set([
```
sonleoracle: `NEVER_TRANSFORM_FIELDS` includes `component_type` and `agentspec_version` but not `componentType` / `agentspecVersion`. If camelCase mode is intended as a full alternate representation, this needs a clearer contract: right now it risks partially-transformed output.

tsagentspec/src/flows/nodes/end-node.ts
```diff
@@ -0,0 +1,35 @@
...
+import { NodeBaseSchema, DEFAULT_NEXT_BRANCH } from "../node.js";
+
+export const EndNodeSchema = NodeBaseSchema.extend({
+  componentType: z.literal("EndNode"),
+  branchName: z.string().default(DEFAULT_NEXT_BRANCH),
+});
+
+export type EndNode = z.infer<typeof EndNodeSchema>;
+
+export function createEndNode(opts: {
```
sonleoracle, line 15: `createEndNode()` uses the same input/output mirroring pattern as `createStartNode()` (`inputs = opts.inputs ?? opts.outputs`, `outputs = opts.outputs ?? opts.inputs`). This can produce serialized configs where EndNode `inputs` incorrectly equal EndNode `outputs` (or vice versa) when only one is provided. That changes the IO surface exposed by the node and can cause Flow output inference / validation to disagree with pyagentspec, where EndNode outputs are the authoritative flow outputs.

tsagentspec/src/flows/flow.ts
```diff
@@ -0,0 +1,93 @@
...
+import { z } from "zod";
+import type { Property } from "../property.js";
+import { ComponentWithIOSchema } from "../component.js";
+import { ControlFlowEdgeSchema, type ControlFlowEdge } from "./edges/control-flow-edge.js";
+import { DataFlowEdgeSchema, type DataFlowEdge } from "./edges/data-flow-edge.js";
+
+// z.record(z.unknown()) is used instead of NodeUnion to break a circular dependency
+// (Flow -> FlowNode -> Flow). Validation of individual nodes happens via their own
+// schemas when constructed through factory functions.
+const NodeRef = z.record(z.unknown());
```
sonleoracle: `NodeRef = z.record(z.unknown())` is used to avoid circular deps, but it means `FlowSchema` itself doesn’t validate that `nodes` are valid Node components. This can allow invalid configs to pass TS parsing and only fail later at serialization/deserialization boundaries.

spichen: Fixed. `node` and `subflow` refs now use `z.lazy()` with late-binding registration to break the circular dep while validating against NodeUnion/FlowSchema at parse time.

tsagentspec/src/flows/nodes/end-node.ts
spichen: Fixed

tsagentspec/src/serialization/serialization-context.ts
spichen: Fixed. Refactored the serialisation logic.

tsagentspec/src/serialization/serialization-context.ts
spichen: Fixed.

tsagentspec/src/serialization/deserialization-context.ts
spichen: Fixed. Refactored the serialization logic.

tsagentspec/src/property.ts
spichen: Fixed.

tsagentspec/src/component.ts
spichen: Fixed.

tsagentspec/src/component.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/** Check if a value is a component (has uuid id, name, componentType) */
+export function isComponent(value: unknown): value is ComponentBase {
+  if (typeof value !== "object" || value === null) return false;
+  const obj = value as Record<string, unknown>;
+  return (
+    typeof obj["id"] === "string" &&
+    UUID_RE.test(obj["id"]) &&
```
sonleoracle: `id` is enforced as UUID, while Agent Spec schema and Python accept generic string IDs, which currently rejects valid fixtures with IDs like `START_NODE`. **Action:** please switch to `z.string()` for `id`, remove UUID regex checking in `isComponent`, and add one regression test with non-UUID IDs.

tsagentspec/src/agents/swarm.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// z.record(z.unknown()) is used instead of AgenticComponentUnion to break a circular
+// dependency (Swarm -> AgenticComponentUnion -> Swarm). Validation of the nested agents
+// is handled by the deserialization plugin at runtime.
+const AgenticComponentRef = z.lazy(() => z.record(z.unknown()));
+
+export const SwarmSchema = ComponentWithIOSchema.extend({
+  componentType: z.literal("Swarm"),
+  firstAgent: AgenticComponentRef,
+  relationships: z.array(z.tuple([AgenticComponentRef, AgenticComponentRef])),
```
sonleoracle: Swarm schema reads cleanly; one semantic parity gap is that empty `relationships` is currently allowed, while Python rejects that shape to prevent invalid swarm setups. **Action:** please add `.min(1)` (or refine) on `relationships` and a negative test for empty arrays.

tsagentspec/src/agents/manager-workers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// z.record(z.unknown()) is used instead of AgenticComponentUnion to break a circular
+// dependency (ManagerWorkers -> AgenticComponentUnion -> ManagerWorkers). Validation of
+// the nested agents is handled by the deserialization plugin at runtime.
+const AgenticComponentRef = z.lazy(() => z.record(z.unknown()));
+
+export const ManagerWorkersSchema = ComponentWithIOSchema.extend({
+  componentType: z.literal("ManagerWorkers"),
+  groupManager: AgenticComponentRef,
+  workers: z.array(AgenticComponentRef),
```
sonleoracle: ManagerWorkers is close to parity; currently `workers` can be empty and there is no guard against `groupManager` also appearing in `workers`. **Action:** please add both validations and corresponding negative tests to match Python behavior.

tsagentspec/src/flows/flow.ts
```diff
@@ -0,0 +1,91 @@
...
+  id?: string;
+  description?: string;
+  metadata?: Record<string, unknown>;
+  inputs?: Property[];
+  outputs?: Property[];
+}): Flow {
+  const inputs = opts.inputs ?? inferFlowInputs(opts.startNode);
+  const outputs = opts.outputs ?? inferFlowOutputs(opts.nodes);
+
+  return Object.freeze(
```
sonleoracle, line 193: `Flow` construction is nicely typed, but unlike Python it currently relies mostly on shape validation and does not enforce key graph invariants (single StartNode, EndNode presence/connectivity, edges referencing nodes in flow). **Action:** please add runtime validations (schema refine or builder-level checks reused in `createFlow`) plus negative tests for malformed graph structures.

tsagentspec/src/flows/edges/data-flow-edge.ts
```diff
@@ -0,0 +1,71 @@
...
+  sourceOutput: string;
+  destinationNode: ComponentWithIO;
+  destinationInput: string;
+  id?: string;
+  description?: string;
+  metadata?: Record<string, unknown>;
+}): DataFlowEdge {
+  const { outputs: sourceOutputs, name: sourceName } = opts.sourceNode;
+  const sourceProperty = findPropertyByTitle(sourceOutputs, opts.sourceOutput);
+  if (sourceOutputs && sourceOutputs.length > 0 && !sourceProperty) {
```
sonleoracle: The edge checks are helpful, though the current conditions skip validation when source/destination IO arrays are empty, which may allow silent invalid wiring that Python rejects. **Action:** please validate referenced IO property names unconditionally (or make the relaxed behavior explicit and tested) and add negative tests where named fields do not exist.

tsagentspec/src/serialization/serialization-context.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    const keys = getProtocolKeys(this.camelCase);
+    const modelDump = this.dumpField(component) as SerializedFields;
+    modelDump[keys.agentspecVersion] = this.agentspecVersion;
+
+    return this.makeOrderedDict(modelDump) as SerializedDict;
+  }
+
+  /** Check if a field should be excluded for the current version */
+  isFieldVersionGated(componentType: string, fieldName: string): boolean {
```
sonleoracle, line 303: The `_self` gate behavior is very close, but right now older target versions can serialize invalid partial components (for example `BuiltinTool` reduced to only `component_type`), which is hard to detect downstream. **Action:** please add a fail-fast version compatibility check (Python-style) before serialization and add a negative test for `BuiltinTool` at `<25.4.2`.

tsagentspec/src/serialization/version-gates.ts
```diff
@@ -0,0 +1,53 @@
...
+ * When serializing for an older version, these fields are excluded.
+ */
+import { AgentSpecVersion } from "../versioning.js";
+import type { ComponentTypeName } from "../component.js";
+
+/**
+ * Maps componentType -> { fieldName -> minimum AgentSpecVersion }
+ * Special key "_self" means the entire component requires that version.
+ */
+export const VERSION_GATED_FIELDS = {
```
sonleoracle, line 14: Nice start on gate coverage; I think a few fields are still missing compared to Python (`OpenAI apiType/apiKey`, `ApiNode/RemoteTool sensitiveHeaders`, remote transport `auth`), which can make back-version exports inaccurate. **Action:** please extend this gate map to match Python `_versioned_model_fields_to_exclude` and add focused tests per gated field.

tsagentspec/src/component.ts
spichen: Done

tsagentspec/src/agents/swarm.ts
spichen: Done

tsagentspec/src/agents/manager-workers.ts
spichen: Done

spichen:
> This is in mergeable shape for me if we keep the support contract narrow.
> 
> I do not want to make this package a broad compatibility gate for every historical fixture in the repo, but I do want one curated repository-fixture round-trip test so we have a stable CI contract for common Agent Spec shapes.
> 
> I locally validated a curated set of `61` configs that round-trip successfully with the current SDK when tested in the right mode:
> 
> * standalone current-format configs are tested as ordinary one-file round trips
> * `howto_disaggregated_*` is tested with the intended two-step API (`importOnlyReferencedComponents` + `componentsRegistry`)
> * the weather/ancestry adapter fixtures are tested from placeholder-resolved copies derived from the repo paths below
> 
> Please add a `tests/repo-fixtures.test.ts`-style test that round-trips these repo configs:
> 
> * `docs/pyagentspec/source/agentspec_config_examples/agentspec_oracle_it_assistant.json`
> * `docs/pyagentspec/source/agentspec_config_examples/agentspec_oracle_it_assistant.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/ancestry_agent_with_client_tool.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/autogen_to_agentspec.json`
> * `docs/pyagentspec/source/agentspec_config_examples/autogen_to_agentspec.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_christmas_greetings_tutorial.json`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_christmas_greetings_tutorial.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_code_assistant_tutorial.json`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_code_assistant_tutorial.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_ops_assistant_tutorial_agent.json`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_ops_assistant_tutorial_agent.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_ops_assistant_tutorial_flow.json`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_ops_assistant_tutorial_flow.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_tutorial_cybersecurity_flow.json`
> * `docs/pyagentspec/source/agentspec_config_examples/ext_tutorial_cybersecurity_flow.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/haiku_without_a_flow.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_ag_ui.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_ag_ui.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_agent_with_remote_tools.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_agent_with_remote_tools.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_agents.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_agents.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_catchexception.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_catchexception.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_disaggregated_component_config.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_disaggregated_component_config.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_disaggregated_main_config.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_disaggregated_main_config.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_flow_with_conditional_branches.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_flow_with_conditional_branches.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_flowbuilder.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_flowbuilder.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_managerworkers.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_managerworkers.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mapnode.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mapnode.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mcp_agent.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mcp_agent.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mcp_flow.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_mcp_flow.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_parallelflownode.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_parallelflownode.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation1.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation1.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation2.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation2.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation3.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_structured_generation3.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_summarization_transforms.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_summary_flow.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_summary_flow.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/math_homework_agent.json`
> * `docs/pyagentspec/source/agentspec_config_examples/math_homework_agent.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/simple_agent_with_rag_tool.json`
> * `docs/pyagentspec/source/agentspec_config_examples/simple_agent_with_rag_tool.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/swarm_calculator.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/weather_agent_client_tool.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/weather_agent_remote_tool.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/weather_agent_server_tool.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/weather_agent_with_outputs.yaml`
> * `pyagentspec/tests/adapters/langgraph/configs/weather_ollama_agent.yaml`
> 
> Notes on that list:
> 
> * the `howto_disaggregated_*` files should be exercised with the intended two-step API, not as naive single-file fixtures
> * the weather/ancestry adapter fixtures should be tested from placeholder-resolved copies derived from those repo paths, not from the raw YAML files with `[[...]]` placeholders
> 
> The broader public-example compatibility follow-ups I would still like to see, but not as merge blockers, are:
> 
> * `docs/pyagentspec/source/agentspec_config_examples/howto_a2aagent.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_a2aagent.yaml`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_swarm.json`
> * `docs/pyagentspec/source/agentspec_config_examples/howto_swarm.yaml`
> 
> One more thing I would like cleaned up before merge:
> 
> * `npm ci` and `npm run build` both succeed locally, but `npm run build` emits a packaging warning because `package.json` declares `exports.types` after `import` and `require`, so that `types` condition is never used
> * `npm audit --omit=dev --json` currently reports a direct runtime vulnerability because `package-lock.json` resolves `yaml` to `2.8.2`; there is a fix available in `>= 2.8.3`
> 
> Please:
> 
> * bump `yaml` to a non-vulnerable version and refresh the lockfile
> * fix the `exports` ordering in `tsagentspec/package.json` so the `types` condition is actually effective
> 
> These are not functional runtime failures in the current branch, but they are package-quality issues that are better fixed before merge.
> 
> Also please briefly document the intended supported surface so we do not imply that every historical/example fixture in this repo is guaranteed to round-trip in TS.

I've added `tests/repo-fixtures.test.ts` with all 61 fixtures - standard files as plain round-trips, howto_disaggregated_* via the two-step API, and the weather/ancestry/haiku files with
  [[...]] tokens swapped in memory before parsing. 

yaml is at 2.8.3 (npm audit --omit=dev is clean), exports.types is before import/require in package.json.

spichen:
@sonleoracle 

- Should this PR also add a CI step to run the new SDK's tests, or handle that separately?
- Should this PR include documentation for the new SDK, or save that for a follow-up?
- Should we move the config files used in repo fixture tests out of the pyagentspec docs into a common shared area?

sonleoracle:
> @sonleoracle
> 
> * Should this PR also add a CI step to run the new SDK's tests, or handle that separately?
> * Should this PR include documentation for the new SDK, or save that for a follow-up?
> * Should we move the config files used in repo fixture tests out of the pyagentspec docs into a common shared area?

Hi @spichen 

This PR is mergeable, and we can handle those in follow-up PRs.

Since we do not currently have plans to publish the TypeScript SDK, I think the installation instructions in the README should be updated to describe source installation instead, for example:

```bash
git clone https://github.com/oracle/agent-spec.git
cd agent-spec/tsagentspec
npm ci
npm run build
```

and, if useful, how to consume it from another local project, e.g. via a local path install.

Related small doc nit: tsagentspec/examples/README.md tells users to run examples with npx tsx / npx ts-node, but those tools do not appear to be declared in devDependencies, so that guidance may also need to be adjusted or clarified.

dhilloulinoracle:
Internal regression failed: Build ID #403

dhilloulinoracle:
Internal regression failed: Build ID #404

dhilloulinoracle:
Internal regression failed: Build ID #405

spichen:
> > @sonleoracle
> > 
> > * Should this PR also add a CI step to run the new SDK's tests, or handle that separately?
> > * Should this PR include documentation for the new SDK, or save that for a follow-up?
> > * Should we move the config files used in repo fixture tests out of the pyagentspec docs into a common shared area?
> 
> Hi @spichen
> 
> This PR is mergeable, and we can handle those in follow-up PRs.
> 
> Since we do not currently have plans to publish the TypeScript SDK, I think the installation instructions in the README should be updated to describe source installation instead, for example:
> 
> ```shell
> git clone https://github.com/oracle/agent-spec.git
> cd agent-spec/tsagentspec
> npm ci
> npm run build
> ```
> 
> and, if useful, how to consume it from another local project, e.g. via a local path install.
> 
> Related small doc nit: tsagentspec/examples/README.md tells users to run examples with npx tsx / npx ts-node, but those tools do not appear to be declared in devDependencies, so that guidance may also need to be adjusted or clarified.

Alright, I've updated the README.
What's the plan for publishing this as an npm package in the future?

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #406

dhilloulinoracle:
> What's the plan for publishing this as an npm package in the future?

We will publish in a more official manner once we have put in place all the required automated checks to ensure that it has the same functionalities as the python one (no divergence in functionalities) as well as passes our internal requirements for security :) .

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #407

tsagentspec/src/component.ts
```diff
@@ -0,0 +1,110 @@
...
+/** ComponentWithIO adds inputs/outputs */
+export const ComponentWithIOSchema = ComponentBaseSchema.extend({
+  inputs: z.array(PropertySchema).optional(),
+  outputs: z.array(PropertySchema).optional(),
+});
+
+export type ComponentWithIO = z.infer<typeof ComponentWithIOSchema>;
+
+/** Check if a value is a component (has string id, name, componentType) */
+export function isComponent(value: unknown): value is ComponentBase {
```
sonleoracle, line 30: This shape-based component detection is too permissive for arbitrary nested values. Any plain object with string `id` + `name` + `componentType` can be treated as a real component even when it appears inside opaque payload fields like `data`, `headers`, `queryParams`, or `metadata`, and the generic recursive deserialization path then resolves nested `$component_ref` values from those payloads as well. That is a regression from `pyagentspec`, which keeps nested payload dictionaries inert unless the schema says the field actually contains a component. Please make component/ref handling field-aware and add regression tests showing that nested payload dictionaries round-trip unchanged.

spichen: Added a static `OPAQUE_FIELDS` set (metadata, data, headers, queryParams, sensitiveHeaders, configuration) gating all three call sites - serialization, deserialization, and referencing traversal.  
Values in those fields are sanitized but never passed through isComponent or isSerializedComponent.

tsagentspec/src/transforms/message-transform.ts
```diff
@@ -0,0 +1,110 @@
...
+  ),
+  summarizedMessageTemplate: z
+    .string()
+    .default("Summarized message: {{summary}}"),
+  maxCacheSize: z.number().int().default(10000),
+  maxCacheLifetime: z.number().int().default(14400),
+  cacheCollectionName: z
+    .string()
+    .default("summarized_messages_cache"),
+  datastore: z.record(z.unknown()).optional(),
```
sonleoracle: pyagentspec does not treat this field as arbitrary JSON: both summarization transforms narrow `datastore` to `SupportedDatastores` (`InMemoryCollectionDatastore | OracleDatabaseDatastore | PostgresDatabaseDatastore`) and validate the schema. TS still accepts any `Record<string, unknown>` here, which means nested `componentType` / `$component_ref` values are still traversed on this field. A concrete repro on this head is:

```javascript
datastore: {
    payload: { id: "nested", name: "nested", componentType: "OpenAiConfig" },
    ref: { $component_ref: "cfg" }
}
```

That serializes `payload` as if it were a real component, and deserialization then fails because `cfg` is treated as a real component ref. Can we narrow this to the actual datastore component union instead of a generic record, and update the `should accept a custom datastore` tests accordingly?

spichen: Done

sonleoracle:
Hi @spichen 
Could you please rebase on top of latest main? We just pushed a few fixes.

spichen:
> Hi @spichen Could you please rebase on top of latest main? We just pushed a few fixes.

I merged main branch into this branch.

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #419

dhilloulinoracle:
Merge Gate failed: Build ID #643

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #446

dhilloulinoracle:
Merge Gate failed: Build ID #644

Issue #101, opened 2026-02-16
docs: missing tracing API reference


dhilloulinoracle:
The API reference doesn't show the tracing APIs, it should be added

PR #102, opened 2026-02-16
Agent Spec Tracing API docs

cesarebernardis:
Closes https://github.com/oracle/agent-spec/issues/101

Add Tracing class documentation to the APIs section of Agent Spec docs.

### DIFF
--- a/docs/pyagentspec/source/_components/all_components.json
+++ b/docs/pyagentspec/source/_components/all_components.json
@@ -298,6 +298,16 @@
         ]
     },
     {
+        "name": "Tracing",
+        "path": "tracing",
+        "classes": [
+            {"path": "pyagentspec.tracing.trace.Trace"},
+            {"path": "pyagentspec.tracing.spanprocessor.SpanProcessor"},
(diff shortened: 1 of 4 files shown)


docs/pyagentspec/source/_components/all_components.json
```diff
@@ -297,6 +297,16 @@
             {"path": "pyagentspec.transforms.ConversationSummarizationTransform"}
         ]
     },
+    {
+        "name": "Tracing",
+        "path": "tracing",
+        "classes": [
+            {"path": "pyagentspec.tracing.trace.Trace"},
+            {"path": "pyagentspec.tracing.spanprocessor.SpanProcessor"},
```
paul-cayet, line 305: You can/should make it shorter with the name so that it renders nicely in the table, something like `pyagentspec.tracing.SpanProcessor` ?
Same for the others

cesarebernardis: I sticked to something that can be used for import, doesn't it make more sense like that?

nasseur:
Internal regression succeeded 🍏: Build ID #409

PR #114, opened 2026-02-19
RFC: Generic LLM Configuration Support

spichen:
Propose a provider-agnostic GenericLlmConfig component that uses a string-based provider.type discriminator and flexible ProviderConfig, replacing the need for per-provider LlmConfig subclasses.

### DIFF
--- a/docs/pyagentspec/source/RFC-generic-llm-config.rst
+++ b/docs/pyagentspec/source/RFC-generic-llm-config.rst
@@ -1,0 +1,686 @@
+:orphan:
+
+==================================
+RFC: Generic LLM Configuration
+==================================
+
+This RFC proposes a ``GenericLlmConfig`` component for
+`Open Agent Spec <https://github.com/oracle/agent-spec>`_ that provides a
+provider-agnostic way to configure LLM connections using a string-based
+provider discriminator and an open-ended provider configuration object.
+
+Summary
+-------
+
+Add a ``GenericLlmConfig`` component that can describe any LLM provider
+through a ``provider.type`` string discriminator and a flexible
+``ProviderConfig`` object, without requiring a new ``LlmConfig`` subclass per
+provider.
+
+Motivation
+----------
+
+The current ``LlmConfig`` hierarchy requires a dedicated subclass for every LLM
+provider (``OpenAiConfig``, ``OciGenAiConfig``, ``OllamaConfig``, etc.).  This
+design creates several practical problems:
+
+1. **Every new provider requires a spec change.**
+   Adding support for a provider today means creating a new ``LlmConfig``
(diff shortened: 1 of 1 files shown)


cesarebernardis:
Hi @spichen, thank you for the great work.

This is a very good starting point, we have been discussing this design internally, and we agreed on a few points.
As mentioned in https://github.com/oracle/agent-spec/pull/108#issuecomment-3921255470, we prefer to make multiple, small, simple changes to the spec, so that it's easier to agree on the design. We would mainly remove the Auth part of it for now, and simplify a bit the current proposal.

Here's what we thought could be a possible solution:

Instead of having a new GenericLlmConfig, we make the LlmConfig not abstract anymore.
We add a few parameters related to providers and apis that are optional strings, type-wise, but in the spec we indicate some known values, similarly to what we do with `api_type` already in some configs.

```python
# Not abstract anymore
class LlmConfig(Component):
    """A generic, provider-agnostic LLM configuration."""

    model_id: str
    """Primary model identifier"""

    provider: Optional[str] = None
    """Model provider (e.g., meta, openai, anthropic, ...)"""

    api_provider: Optional[str] = None
    """APIs provider (e.g., oci, openai, vertex_ai, aws_bedrock, ...)"""

    api_type: Optional[str] = None
    """API protocol to use (e.g., chat_completions, responses, ...)"""
```

Existing classes will fix the value of these attributes where needed, and they should be excluded from the serialization for brevity.

```python
class OciGenAiConfig(GenericLlmConfig):
    # Don't know the (model) provider a priori
    api_provider: str = "oci"  # Freeze the value
    ...

class OpenAiConfig(GenericLlmConfig):
    # We know both provider and api_provider, we fix them
    provider: str = "openai"  # Freeze the value 
    api_provider: str = "openai"  # Freeze the value 
    ...

class OpenAiCompatibleConfig(GenericLlmConfig):
    # This limits only the api_type to chat_completions or responses
    # We don't know what is the model, nor the api provider
    ...
```

The name of the attributes are chosen to fit what is already present in some LlmConfig classes (you can give a look at the OCI ones for example).

The main reason for this choice is that all the attributes are quite independent from each other, and collecting them in a Provider class would require to fix one of the attributes, but in an unclear manner, and the number of subclasses could explode. For example: meta model exposed by AWS bedrock behind responses APIs, cohere model exposed by OCI behind chat_completions, ...
We think it would be a wiser start to have something (almost) free form and not encapsulated in classes, so that users are free to support new models, providers, and APIs without waiting for a new version of Agent Spec, which is I think the biggest value of this proposal.

Let us know what you think about it.

cesarebernardis:
Merged as part of https://github.com/oracle/agent-spec/pull/147

PR #118, opened 2026-02-23
feat: add timeout field to RemoteTool

MAabdelmoumen:
Add an optional `timeout` field to `RemoteTool` that is forwarded directly to `httpx.request()`. This allows agent spec authors to configure per-tool HTTP timeouts for slow endpoints such as transcription or inference services.

The field defaults to `None`, preserving existing behaviour where httpx uses its own default timeout (5 seconds).

Example usage in an agent spec:
  {
    "component_type": "RemoteTool",
    "timeout": 300.0,
    ...
  }


dhilloulinoracle:
Hi @MAabdelmoumen ,

We recently merged https://github.com/oracle/agent-spec/pull/121 bring configurations for this kind of things.
Would it be possible for you to have a look and tell us if that seems to add the capabilities you wanted?

PR #147, opened 2026-03-24
Generic LLM Configuration Support

spichen:
Implementation of RFC https://github.com/oracle/agent-spec/pull/114


docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -530,14 +530,26 @@ In order to make them usable, we need to let users specify all the
...
 
 .. code-block:: python
 
    class LlmConfig(Component):
+     model_id: str
+     provider: Optional[str]
+     api_provider: Optional[str]
+     api_type: Optional[str]
+     url: Optional[str]
+     api_key: Optional[str]
```
cesarebernardis: Please mark the `api_key` as sensitive
```suggestion
     api_key: SensitiveField[Optional[str]]
```

There's also a table at the end of this file with all the sensitive fields, please that too

docs/pyagentspec/source/misc/reference_sheet.rst
```diff
@@ -25,16 +25,20 @@ Agent Spec defines a Component called ``LlmConfig`` that contains all necessary
 .. code-block:: python
 
    class LlmConfig(Component):
+     model_id: str
+     provider: Optional[str]
+     api_provider: Optional[str]
+     api_type: Optional[str]
```
cesarebernardis, line 31: Some fields are missing, please align the definition with the one in the spec

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -530,14 +530,26 @@ In order to make them usable, we need to let users specify all the
...
      default_generation_parameters: Optional[Dict[str, Any]]
 
-We require only to specify the default generation parameters that should be used by default when prompting the LLM.
+The ``model_id`` field is required and identifies the model to use.
+The ``provider`` field is optional and identifies the model provider (e.g. ``"openai"``, ``"meta"``, ``"anthropic"``, ``"cohere"``).
+The ``api_provider`` field is optional and identifies the API provider serving the model (e.g. ``"openai"``, ``"oci"``, ``"vllm"``, ``"ollama"``, ``"aws_bedrock"``, ``"vertex_ai"``).
+The ``api_type`` field is optional and identifies the wire protocol to use (e.g. ``"chat_completions"``, ``"responses"``).
+The ``url`` field is optional and specifies the URL of the API endpoint (e.g. ``"https://api.openai.com/v1"``).
```
cesarebernardis: I would maybe specify that, if null, the default API url of the API provider (if any) should be used

pyagentspec/src/pyagentspec/component.py
```diff
@@ -467,6 +467,10 @@ def model_json_schema(
             all_subclasses = cls._get_all_subclasses(only_core_components=only_core_components)
             adapter = TypeAdapter(Union[all_subclasses])  # type: ignore
             json_schema = adapter.json_schema(by_alias=by_alias, mode=mode)
+        elif getattr(cls, "_include_subclasses_in_schema", False):
```
cesarebernardis: Could you help understanding why you need this new logic?

pyagentspec/src/pyagentspec/component.py
```diff
@@ -572,6 +576,30 @@ def get_validation_errors(
...
+    json_schema: JsonSchemaValue,
+    subclasses: list[type],
+    mode: JsonSchemaMode,
+    by_alias: bool,
+) -> None:
+    """Add missing subclass schemas to ``$defs``.
+
+    When a concrete class also serves as a base (e.g. ``LlmConfig``), Pydantic
+    only generates the parent schema. Phase 1 (abstract resolution) skips it
+    because it is not abstract, so subclass schemas are never added to ``$defs``.
```
cesarebernardis: Could you double check this? Because I think we have this pattern already in the class tree (e.g., `OllamaConfig` is child of `OpenAiCompatibleConfig`, which is not abstract, and it shows up in the json schema).

Maybe I misunderstood the issue?

Maybe the problem raises when the starting class form which we call `model_json_schema` is not abstract?

pyagentspec/tests/test_bare_llmconfig_adapter_dispatch.py
```diff
@@ -0,0 +1,142 @@
...
+def bare_llmconfig_openai_responses() -> LlmConfig:
+    return LlmConfig(name="test", model_id="gpt-4o", api_provider="openai", api_type="responses")
+
+
+@pytest.fixture
+def bare_llmconfig_unsupported() -> LlmConfig:
+    return LlmConfig(name="test", model_id="some-model", api_provider="unsupported_provider")
+
+
+class TestOpenAiAgentsDispatch:
```
cesarebernardis: For tests that are adapter-specific we have ad-hoc folders in `tests/adapters`, and tests in that folder are skipped automatically if the extra dependencies required are not installed.

Please split these tests in the respective adapter folders.

pyagentspec/src/pyagentspec/component.py
spichen: `model_json_schema()` used to hit the `cls._is_abstract` branch, which builds a `Union[all_subclasses]`. Now that the class is concrete, it falls through to the `else` branch and produces only `LlmConfig`'s own schema - the subclasses are missing entirely. `_include_subclasses_in_schema` adds a third path that builds `Union[cls, *all_subclasses]`, so the full hierarchy still appears in the output.

pyagentspec/src/pyagentspec/component.py
spichen: Yes, it is because of the starting class now not being abstract. `OpenAiCompatibleConfig` case worked earlier because the starting class was abstract `Llmconfig`. I added a test case to cover this.

Do you better approach to solve this?

pyagentspec/src/pyagentspec/component.py
cesarebernardis: I see, I think this is a more generic bug in `pyagentspec` then. I would recommend to go over subclasses in any case, independently from abstract or non-abstract class, which should only determine if the class itself should be included or not. Something like

```
all_subclasses = cls._get_all_subclasses(only_core_components=only_core_components)
adapter = TypeAdapter(Union[tuple([*all_subclasses] if cls._is_abstract else [cls, *all_subclasses])])  # type: ignore
json_schema = adapter.json_schema(by_alias=by_alias, mode=mode, **kwargs)
```

instead of the existing if-else.

I would expect to not need the new `_ensure_subclass_schemas_exist` anymore as well.

Note also that some tests might start failing after this change, because this will start including children of concrete classes in json schemas. You should be able to make all failing the tests green again by limiting the scope of model_json_schema to `only_core_components=True` in the tests that need it. If you don't notice this issue, no worries, I'll patch it when running the internal regressions.

pyagentspec/src/pyagentspec/component.py
cesarebernardis: Yes, I think that after addressing https://github.com/oracle/agent-spec/pull/147#discussion_r3022358188 this should not be needed anymore

pyagentspec/src/pyagentspec/component.py
spichen: I made this change now. Much cleaner 👍

cesarebernardis (approved):

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -530,15 +530,27 @@ In order to make them usable, we need to let users specify all the
...
+     url: Optional[str]
+     api_key: SensitiveField[Optional[str]]
      default_generation_parameters: Optional[Dict[str, Any]]
      retry_policy: Optional[RetryPolicy]
 
-We require only to specify the default generation parameters that should be used by default when prompting the LLM.
+The ``model_id`` field is required and identifies the model to use.
+The ``provider`` field is optional and identifies the model provider (e.g. ``"openai"``, ``"meta"``, ``"anthropic"``, ``"cohere"``).
+The ``api_provider`` field is optional and identifies the API provider serving the model (e.g. ``"openai"``, ``"oci"``, ``"vllm"``, ``"ollama"``, ``"aws_bedrock"``, ``"vertex_ai"``).
+The ``api_type`` field is optional and identifies the wire protocol to use (e.g. ``"chat_completions"``, ``"responses"``).
```
sonleoracle, line 550: "wire protocol" is not super clear, suggestion: API format

docs/pyagentspec/source/howtoguides/howto_llm_from_different_providers.rst
```diff
@@ -34,6 +36,51 @@ For example, you can attach a retry policy directly to a ``VllmConfig``:
...
+
+  The model provider, i.e. who made the model (e.g. ``"openai"``, ``"meta"``, ``"anthropic"``, ``"cohere"``).
+
+.. option:: api_provider: str, null
+
+  The API provider, i.e. who serves the API (e.g. ``"openai"``, ``"oci"``, ``"vllm"``, ``"ollama"``, ``"aws_bedrock"``, ``"vertex_ai"``).
+
+.. option:: api_type: str, null
+
+  The wire protocol to use (e.g. ``"chat_completions"``, ``"responses"``).
```
sonleoracle:
```suggestion
  The API format to use to interact with the LLM (e.g. ``"chat_completions"``, ``"responses"``).
```

pyagentspec/src/pyagentspec/llms/geminiconfig.py
```diff
@@ -23,7 +23,16 @@ class GeminiConfig(LlmConfig):
         exclude=True,
     )
 
-    model_id: str
-    """Identifier of the Gemini model to use."""
     auth: SerializeAsAny[GeminiAuthConfig]
     """Authentication configuration used to connect to the Gemini service."""
+    provider: str = "google"
```
sonleoracle, line 28: For GeminiConfig specifically, its current behavior is that it supports two API providers: Google AI Studio and GCP Vertex AI, and this is currently determined by which GeminiAuthConfig is passed to the `auth` object. Perhaps we should also override the `api_provider` here too, and the adapters would need to align `api_provider` with the `auth` field?

pyagentspec/src/pyagentspec/llms/llmconfig.py
```diff
@@ -11,17 +11,38 @@
...
 
-    This class provides the base class, while concrete classes provide the required
-    configuration to connect to specific LLM providers.
+    This class can be used directly with the ``provider``, ``api_provider``, and ``api_type``
+    fields to describe any LLM without a dedicated subclass. Concrete subclasses provide
+    additional configuration for specific LLM providers.
     """
 
+    model_id: str
+    """ID of the model to use"""
```
sonleoracle, line 28: Any particular guidance on the model id string? e.g. should it match exactly what is served by the API provider?

pyagentspec/src/pyagentspec/adapters/autogen/_autogenconverter.py
```diff
@@ -209,8 +209,17 @@ def _prepare_llm_args(
...
             return AutogenOpenAIChatCompletionClient(**_prepare_llm_args(agentspec_llm))
         else:
+            # Bare LlmConfig — dispatch on api_provider string
+            if agentspec_llm.api_provider == "openai":
+                kwargs: Dict[str, Any] = {"model": agentspec_llm.model_id}
+                if agentspec_llm.url is not None:
+                    kwargs["base_url"] = agentspec_llm.url
+                if agentspec_llm.api_key is not None:
+                    kwargs["api_key"] = agentspec_llm.api_key
+                return AutogenOpenAIChatCompletionClient(**kwargs)
```
sonleoracle, line 219: please ensure this is equivalent to the `AgentSpecOpenAiCompatibleModel` branch in line 209 just above, right now it is missing at least `model_info`

pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py
```diff
@@ -1306,8 +1306,19 @@ def _llm_convert_to_langgraph(
                 **self._oci_client_config_to_langgraph(llm_config.client_config),
             )
         else:
+            # Bare LlmConfig — dispatch on api_provider string
+            if llm_config.api_provider == "openai":
+                return _create_chat_openai_model(
+                    model_id=llm_config.model_id,
+                    base_url=llm_config.url,
```
sonleoracle, line 1313: Currently we have normalization for user-supplied base_urls so that we ensure the final url is correct. Here, I notice that for the new LlmConfig.url, there is no normalization being applied in any adapters, so migrating existing Agent Spec configs might be a tricky because of the URL

docs/pyagentspec/source/misc/reference_sheet.rst
```diff
@@ -25,16 +25,22 @@ Agent Spec defines a Component called ``LlmConfig`` that contains all necessary
 .. code-block:: python
 
    class LlmConfig(Component):
+     model_id: str
```
sonleoracle, line 28: Non-blocking: it looks like the new generic LLM fields are intended for bare LlmConfig, not for provider specific subclasses like OpenAiConfig/VllmConfig. The problem is that those subclasses still inherit these fields as valid constructor inputs, so something like OpenAiConfig(provider="meta", api_provider="foo") is accepted, but those values are then dropped during serialization and replaced by the subclass defaults on deserialize.

If that usage is unsupported, could we enforce it explicitly at validation time (or otherwise prevent overriding these inherited fields on subclasses)? As-is, the API accepts invalid input but round-trips it lossy, which is likely to surprise users.

sonleoracle (approved):

pyagentspec/src/pyagentspec/llms/geminiconfig.py
spichen: Agree, `api_provider` field should be specifying this.

cesarebernardis: I think that it's ok as-is right now, as `api_provider` and `provider` are two different things. While `api_provider` can be different based on the `auth`, the (model) `provider` in Gemini models is always `google`.

pyagentspec/src/pyagentspec/llms/llmconfig.py
cesarebernardis: Yes, maybe we can make it clearer in the spec

xdave:
Hi there, I am eagerly awaiting the incorporation of these changes. It cleanly unlocks a lot wider use of this standard by many individuals and organizations. Thank you!

dhilloulinoracle (approved):

docs/pyagentspec/source/misc/reference_sheet.rst
cesarebernardis: Done

pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py
cesarebernardis: DOne

dhilloulinoracle:
Internal regression failed: Build ID #427

dhilloulinoracle:
Internal regression failed: Build ID #428

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #434

dhilloulinoracle:
Merge Gate failed: Build ID #652

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #451

dhilloulinoracle:
Merge Gate failed: Build ID #653

dhilloulinoracle:
Merge Gate failed: Build ID #654

dhilloulinoracle:
Merge Gate failed: Build ID #655

dhilloulinoracle:
Merge Gate failed: Build ID #656

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #452

dhilloulinoracle:
Merge Gate failed: Build ID #657

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #453

dhilloulinoracle:
Merge Gate failed: Build ID #658

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #454

dhilloulinoracle:
Merge Gate failed: Build ID #659

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #455

dhilloulinoracle:
Merge Gate failed: Build ID #660

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #456

dhilloulinoracle:
Merge Gate failed: Build ID #661

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #457

dhilloulinoracle:
Merge Gate failed: Build ID #662

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #458

dhilloulinoracle:
Merge Gate failed: Build ID #663

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #459

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #664

PR #150, opened 2026-04-08
fix: remove output schema restriction for tools with requires_confirmation

### LINKED ISSUE #149: LangGraph adapter rejects typed outputs on tools with requires_confirmation
## What happens

When a tool has `requires_confirmation=True` and typed outputs, `_ensure_checkpointer_and_valid_tool_config` throws:                                                                                                                    

```                                                                                                                                                                                                                               
Invalid output schema for tool 'tool-name' requiring tool confirmation: json schema should be                                                                                                                                                
left unspecified when using tool confirmation, was [Property(...)].                                                                                                                                                                     
Please use outputs=[Property(title="bash", json_schema={})]                                                                                                                                                                             
```

## Why the check isn't needed                                                                                                                                                                                                           
                                                                                                                                                                                                                                          
`_confirm_then` returns the original typed output on approval and a plain string on rejection. Both are valid LangChain tool message content -LangGraph doesn't actually enforce output schemas at runtime. They're hints for the LLM, not validation contracts.                                                                                                                                                                                                               

There's a TODO in the code that hints at this:                                                                                                                                                                                          

```python
  # TODO: refine to only raise output property does not support string
```

spichen:
Resolves https://github.com/oracle/agent-spec/issues/149

Tools with `requires_confirmation=True` and typed outputs get rejected by `_ensure_checkpointer_and_valid_tool_config`. This forces consumers to strip output schemas before loading, which loses information the LLM could use.

The check exists because `_confirm_then` returns a plain string on rejection, and the worry was it would conflict with a typed schema. It doesn't — `ToolMessage.content` is always a string in LangChain. The schema is never validated at runtime. Both the approved path (normal output) and rejected path (denial string) work fine.

The code already had a TODO acknowledging this was too strict:

```python
# TODO: refine to only raise output property does not support string
```

This PR removes the output schema validation entirely (keeps the checkpointer check) and adds a test with typed outputs to prove it works.

### DIFF
--- a/pyagentspec/tests/adapters/langgraph/test_tools.py
+++ b/pyagentspec/tests/adapters/langgraph/test_tools.py
@@ -837,32 +837,42 @@ async def test_async_server_tool_in_agent_executes_via_ainvoke() -> None:
     assert "10" in str(tool_result_message.content)
 
 
-def test_server_tool_confirmation_with_typed_outputs_works() -> None:
-    """Tools with requires_confirmation and typed output schemas should load
-    without error.  The output schema is metadata for the LLM, not a runtime
-    validation constraint, so it should not be rejected."""
+def test_server_tool_confirmation_with_typed_object_output_works() -> None:
+    """Tools with requires_confirmation and a typed (object) output schema
+    should load without error and execute the approved path. The output schema
+    is metadata for the LLM, not a runtime constraint, and on rejection the
+    denial string maps cleanly into a single declared output."""
     from langchain_core.runnables import RunnableConfig
     from langgraph.checkpoint.memory import MemorySaver
 
     from pyagentspec.adapters.langgraph import AgentSpecLoader
 
+    bash_result = {"stdout": "hello", "stderr": "", "exit_code": 0}
+
     def bash_func(command: str) -> dict:
-        return {"stdout": "hello", "stderr": "", "exit_code": 0}
+        return bash_result
 
     server_tool = ServerTool(
         name="bash",
         description="Run a shell command",
         inputs=[Property(title="command", json_schema={"title": "command", "type": "string"})],
         outputs=[
-            Property(title="stdout", json_schema={"title": "stdout", "type": "string"}),
-            Property(title="stderr", json_schema={"title": "stderr", "type": "string"}),
-            Property(title="exit_code", json_schema={"title": "exit_code", "type": "number"}),
+            Property(
+                title="result",
+                json_schema={
+                    "type": "object",
+                    "properties": {
+                        "stdout": {"type": "string"},
+                        "stderr": {"type": "string"},
+                        "exit_code": {"type": "number"},
+                    },
+                },
+            ),
         ],
         requires_confirmation=True,
     )
     flow = _make_simple_flow_with_tool(ToolNode(name="bash_node", tool=server_tool))
 
-    # Should not raise ValueError about output schema
     app = AgentSpecLoader(
         tool_registry={"bash": bash_func},
         checkpointer=MemorySaver(),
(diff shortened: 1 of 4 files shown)


dhilloulinoracle:
@sonleoracle

sonleoracle:
@spichen thank you for your contribution, this is a nice PR :)

Could you please address the formatting issues and rebase?

spichen:
> @spichen thank you for your contribution, this is a nice PR :)
> 
> Could you please address the formatting issues and rebase?

Done. Could you run checks now please?

pyagentspec/tests/adapters/langgraph/test_tools.py
```diff
@@ -786,6 +786,48 @@ async def double_tool_func(x: int) -> int:
...
+        description="Run a shell command",
+        inputs=[Property(title="command", json_schema={"title": "command", "type": "string"})],
+        outputs=[
+            Property(title="stdout", json_schema={"title": "stdout", "type": "string"}),
+            Property(title="stderr", json_schema={"title": "stderr", "type": "string"}),
+            Property(title="exit_code", json_schema={"title": "exit_code", "type": "number"}),
+        ],
+        requires_confirmation=True,
+    )
+    flow = _make_simple_flow_with_tool(ToolNode(name="bash_node", tool=server_tool))
```
sonleoracle, line 894: This test is currently failing because `_make_simple_flow_with_tool()` only wires the first output. I would suggest to either:
- use a single typed object output and assert exact output, or
- manually build a flow with all three output edges wired.

sonleoracle (changes requested):
Thanks for the PR. I agree the current validation is too strict: `requires_confirmation=True` should not force consumers to erase useful typed output schemas.

I don’t think we should remove the check entirely, though. The tricky case is a Flow `ToolNode` with multiple outputs. On approval, a tool can return a dict like:

```python
{"stdout": "hello", "stderr": "", "exit_code": 0}
```

and that maps cleanly. But on rejection, `_confirm_then()` returns a single denial string. There is no principled way to map that string into multiple declared outputs like stdout, stderr, and exit_code.

In WayFlow, Agent Spec's reference runtime, its behavior is conservative here: it allows typed/multiple outputs, but for rejected confirmed tools in flows it raises by default because rejection may break structured flow outputs. It does not synthesize a fake multi output rejection dict.

So I think the fix should be:
 
1. Allow typed output schemas.
2. Keep approved multi-output tools working when they return a dict.
3. Make rejection for multi-output Flow ToolNodes fail explicitly with a clear error.

That preserves the useful part of this PR without leaving rejection semantics undefined.

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #462

cesarebernardis:
> Thanks for the PR. I agree the current validation is too strict: `requires_confirmation=True` should not force consumers to erase useful typed output schemas.
> 
> I don’t think we should remove the check entirely, though. The tricky case is a Flow `ToolNode` with multiple outputs. On approval, a tool can return a dict like:
> 
> ```python
> {"stdout": "hello", "stderr": "", "exit_code": 0}
> ```
> 
> and that maps cleanly. But on rejection, `_confirm_then()` returns a single denial string. There is no principled way to map that string into multiple declared outputs like stdout, stderr, and exit_code.
> 
> In WayFlow, Agent Spec's reference runtime, its behavior is conservative here: it allows typed/multiple outputs, but for rejected confirmed tools in flows it raises by default because rejection may break structured flow outputs. It does not synthesize a fake multi output rejection dict.
> 
> So I think the fix should be:
> 
> 1. Allow typed output schemas.
> 2. Keep approved multi-output tools working when they return a dict.
> 3. Make rejection for multi-output Flow ToolNodes fail explicitly with a clear error.
> 
> That preserves the useful part of this PR without leaving rejection semantics undefined.

I agree, that's a misbehavior in the current version of the langgraph converter. 

Both 
- the single output of type object with three properties
- the three output properties

should work with the required confirmation flag. The problem right now is in the fact that the `_confirm_then` wrapper returns a string when confirmation is denied (breaking the tool's output structure contract), while it should raise an error instead. We should make it raise.

sonleoracle:
@spichen please ensure the static checks are passing and rebase onto latest main

dhilloulinoracle:
Internal regression failed: Build ID #487

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #488

sonleoracle (approved):
LGTM!

dhilloulinoracle (approved):

dhilloulinoracle:
Internal regression failed: Build ID #489

dhilloulinoracle:
Internal regression failed: Build ID #492

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #496

sonleoracle (changes requested):
Hi @spichen, after some thought, I don't think the raise_on_denial flag is needed and I think we should always raise whenever the tool confirmation is denied. That way, we can keep relying on self.convert in tool_node_convert_to_langgraph so that it covers all tool types and populates converted_components correctly.

Could you please update the code so that we always raise on denial?

cesarebernardis (approved):

dhilloulinoracle:
Internal regression failed: Build ID #513

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #514

sonleoracle (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #507

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #720

PR #163, opened 2026-04-21
Forward LangChain agent middleware from AgentSpecLoader to create_agent

### LINKED ISSUE #162: LangGraph adapter: forward middleware to `create_agent`
The LangGraph `AgentSpecLoader` doesn't let you pass middleware into the `langchain_agents.create_agent(...)` call that happens inside `_create_react_agent_with_given_info`. That's the problem — `create_agent` already takes a `middleware=` list (HITL, state rewriters, etc.), but to inject one right now you have to subclass the converter and copy-paste the whole ~175-line method just to add that one keyword.

spichen:
Add a `middleware` keyword on the LangGraph `AgentSpecLoader` and its converter. When the list is non-empty it's passed as `middleware=` to `langchain_agents.create_agent` inside `_create_react_agent_with_given_info`. Both the loader and the converter copy the list on construction, so later mutations by the caller don't leak in.

If `middleware` is `None` or `[]`, `create_agent` is called exactly as before, no `middleware=` kwarg and no behavior change for existing users.

Why: downstream code that wants to use its own state/message-rewriting middleware currently has to subclass the converter and copy-paste the entire `_create_react_agent_with_given_info` method just to thread the keyword through. Middleware is a LangGraph thing, so this parameter only shows up on the LangGraph adapter. Other adapters have their own loader classes and aren't affected.

Resolves #162


dhilloulinoracle:
Thanks @spichen for this contribution !
The change makes sense to me, so let's get it in :) .

@sonleoracle , please have a look

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,155 @@
...
+    from langchain_core.language_models.fake_chat_models import FakeMessagesListChatModel
+    from langchain_openai import ChatOpenAI
+
+    class FakeModel(FakeMessagesListChatModel, ChatOpenAI):
+        pass
+
+    return FakeModel(responses=[AIMessage(content="Done")])
+
+
+def _build_agent() -> Agent:
```
sonleoracle: I think this could be a pytest fixture to follow the existing testing patterns

pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py
sonleoracle, line 739: Could you please add the middleware here too, and in the AgentNodeExecutor, so that agents inside flows can use middleware? Please also add a regression test for this, and another test that executes the middleware (can be monkey-patched if needed)

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,155 @@
...
+        AgentSpecToLangGraphConverter,
+    )
+    from pyagentspec.adapters.langgraph._types import langchain_agents
+
+    captured: Dict[str, Any] = {}
+    agent_spec = _build_agent()
+    with patch.object(
+        AgentSpecToLangGraphConverter,
+        "_llm_convert_to_langgraph",
+        return_value=_get_fake_model(),
```
sonleoracle: I don't think we need `_get_fake_model()` here, it's enough to pass `return_value=object()`.

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,155 @@
...
+from langchain_core.messages import AIMessage
+
+from pyagentspec.adapters.langgraph import AgentSpecLoader
+from pyagentspec.agent import Agent
+from pyagentspec.llms import OpenAiCompatibleConfig
+from pyagentspec.property import Property
+from pyagentspec.tools import ClientTool
+
+
+def _get_fake_model() -> Any:
```
sonleoracle: I think there is already an existing helper `_get_fake_model` in `test_tools.py`, please reuse that one if you need a fake chat model

sonleoracle (changes requested):

sonleoracle (changes requested):
@spichen
Thanks for the contribution!
could you please resolve the formatting issues and my comments? I think this is a nice contribution to the repo :)

dhilloulinoracle:
Internal regression failed: Build ID #461

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #468

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -49,6 +49,13 @@ class AgentSpecLoader(AdapterAgnosticAgentSpecLoader):
         type names match only the exact serialized component type. When allow and
         block entries both match, the closest match in the component class hierarchy
         wins; block entries win same-distance ties.
+    middleware:
+        Optional list of LangChain agent middleware instances forwarded verbatim to
+        ``langchain_agents.create_agent(middleware=...)`` when compiling an Agent Spec
+        ``Agent`` into a ReAct graph. Order is preserved — index ``0`` is the outermost
+        middleware. When ``None`` or an empty list, the ``middleware`` keyword is
+        omitted entirely from the ``create_agent`` call and behavior is identical to
```
sonleoracle: I don't think we need to mention "and behavior is identical to earlier releases."

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+        ), patch.object(langchain_agents, "create_agent", side_effect=_spy_create_agent(captured)):
+            loader_or_converter = loader_factory(MemorySaver())
+            with pytest.raises(_StopCreateAgent):
+                loader_or_converter.load_component(agent_spec)
+        return captured
+
+    return _capture
+
+
+def test_default_omits_middleware_kwarg(
```
sonleoracle: Can these two tests be parametrized? `test_default_omits_middleware_kwarg` and `test_empty_list_omits_middleware_kwarg`

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+        return loader
+
+    captured = capture_create_agent_kwargs(make_loader)
+    assert captured.get("middleware") == [a]
+
+
+def test_middleware_forwarded_through_flow_agent_node(agent_flow: Flow) -> None:
+    """Regression: middleware must reach ``create_agent`` for agents inside flows.
+
+    Before this fix, ``AgentNodeExecutor`` instantiated a fresh
```
sonleoracle: I don't think it's necessary to mention this detail here

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        from pyagentspec.adapters.langgraph._types import langchain_agents
+
+        captured: Dict[str, Any] = {}
+        with patch.object(
+            AgentSpecToLangGraphConverter,
+            "_llm_convert_to_langgraph",
+            return_value=object(),
+        ), patch.object(langchain_agents, "create_agent", side_effect=_spy_create_agent(captured)):
+            loader_or_converter = loader_factory(MemorySaver())
```
sonleoracle: nit: _capture unconditionally calls `.load_component(...)` so it's always the loader, not the converter

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+) -> None:
+    """A non-empty list reaches ``create_agent`` in the original order."""
+    a, b = object(), object()
+    captured = capture_create_agent_kwargs(
+        lambda cp: AgentSpecLoader(checkpointer=cp, middleware=[a, b])
+    )
+    assert captured.get("middleware") == [a, b]
+
+
+def test_converter_accepts_middleware_directly(agent_spec: Agent) -> None:
```
sonleoracle: Nit: I don't think this test is necessary since users are not expected to use `AgentSpecToLangGraphConverter` directly, and the test `test_middleware_forwarded_in_order` test should already cover this

sonleoracle (approved):

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+from typing import Any, Callable, Dict, List
+from unittest.mock import patch
+
+import pytest
+
+from pyagentspec.adapters.langgraph import AgentSpecLoader
```
cesarebernardis: We should lazily import this to make sure that tests do not fail if the langgraph extra dependency is not installed. You can check what is done in other langgraph adapter tests.

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+
+    fake_model = make_fake_chat_model(responses=[AIMessage(content="Done")])
+
+    calls: List[str] = []
+
+    class _RecordingMiddleware(AgentMiddleware):
+        def before_agent(self, state: Any, runtime: Any) -> None:  # type: ignore[override]
+            calls.append("before_agent")
+
+    with patch.object(
```
cesarebernardis: Instead of monkey patching `_llm_convert_to_langgraph` everywhere, you can use the components registry to pass the model to use

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
```diff
@@ -0,0 +1,247 @@
...
+def _spy_create_agent(captured: Dict[str, Any]) -> Callable[..., Any]:
+    def spy(**kwargs: Any) -> Any:
+        captured.update(kwargs)
+        raise _StopCreateAgent()
+
+    return spy
+
+
+@pytest.fixture
+def agent_spec() -> Agent:
```
cesarebernardis: `agent_spec` is a bit confusing as a name, maybe just call it `agent`, or `agent_with_client_tool`, or something similar

pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py
```diff
@@ -188,6 +188,22 @@ def _exec_body(ns: Dict[str, Any]) -> None:
 
 
 class AgentSpecToLangGraphConverter:
+    def __init__(self, middleware: Optional[List[Any]] = None) -> None:
```
cesarebernardis: Let's keep the same behavior we have for other similar langgraph/langchain-specific parameters, i.e., do not put it in the `init`, but in the `convert` method(s).

pyagentspec/tests/adapters/langgraph/test_middleware_parameter.py
spichen: Done everywhere except test_middleware_list_is_copied - that one goes through AgentSpecLoader.load_component(), which doesn't expose converted_components. Kept the patch rather than extending the loader API in this PR.

sonleoracle:
@spichen thanks for the update.  I notice that middleware is forwarded to direct `AgentNode`s in a `Flow`, but not through nested subflows from `FlowNode`, `MapNode`, or `CatchExceptionNode`. Those helpers cal `self.convert(subflow, ...)` without passing `middleware=middleware`, so agents inside those nested subflows will not receive it.

I think it would be great if this PR also recursively forwards the middleware through all these.

dhilloulinoracle:
Internal regression failed: Build ID #493

dhilloulinoracle:
Internal regression failed: Build ID #495

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #497

pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py
```diff
@@ -190,16 +190,41 @@ def _exec_body(ns: Dict[str, Any]) -> None:
 
 
 class AgentSpecToLangGraphConverter:
+    def __init__(self) -> None:
```
cesarebernardis: Leftover?

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -60,6 +66,7 @@ def __init__(
         *,
         allowed_components: Optional[ComponentPolicyInput] = None,
         blocked_components: Optional[ComponentPolicyInput] = None,
+        middleware: Optional[List[Any]] = None,
```
cesarebernardis: Nit: maybe move it close to `config` for consistency with other APIs. Also, the `*` parameter looks in an arbitrary place, we can remove it

cesarebernardis (approved):

dhilloulinoracle (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #502

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #504

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #716

PR #175, opened 2026-05-05
Use langchain tool inputs parameter in LangGraph tracing

cesarebernardis:
This updates LangGraph tool tracing to normalize callback inputs and use the LangChain inputs parameter by default, so traced tool requests preserve the original structured arguments more reliably.

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
@@ -571,6 +571,7 @@ class AgentSpecToolCallbackHandler(AgentSpecCallbackHandler):
         request_event = AgentSpecToolExecutionRequest(
             request_id=run_id_str,
             tool=self.tool,
+            # LangChain should provide structured `inputs` in the `kwargs`, the others are fallback options
             inputs=_normalize_tool_inputs(self.tool, input_str, kwargs.get("inputs")),
         )
         # starting a tool span for this tool
@@ -667,6 +668,7 @@ class AgentSpecToolCallbackHandler(AgentSpecCallbackHandler):
         request_event = AgentSpecToolExecutionRequest(
             request_id=run_id_str,
             tool=self.tool,
+            # LangChain should provide structured `inputs` in the `kwargs`, the others are fallback options
             inputs=_normalize_tool_inputs(self.tool, input_str, kwargs.get("inputs")),
         )
         # starting a tool span for this tool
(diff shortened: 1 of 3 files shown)


dhilloulinoracle (approved):

dhilloulinoracle:
please add a changelog entry related to the improvement

pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
```diff
@@ -552,7 +571,7 @@ def on_tool_start(
         request_event = AgentSpecToolExecutionRequest(
             request_id=run_id_str,
             tool=self.tool,
-            inputs=ast.literal_eval(input_str) if isinstance(input_str, str) else input_str,
+            inputs=_normalize_tool_inputs(self.tool, input_str, kwargs.get("inputs")),
```
sonleoracle, line 575: when do we need to consider the kwargs of on_tool_start callback? `kwargs.get("inputs")`

cesarebernardis: LangChain provides it, that should be the desired path. The others are fallbacks for older versions/custom incomplete implementations. I'll add a comment

pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
```diff
@@ -75,6 +74,26 @@
...
+    if isinstance(input_value, dict):
+        return input_value
+
+    if tool.inputs and len(tool.inputs) == 1:
+        return {tool.inputs[0].title: input_value}
+
+    if isinstance(input_value, str):
+        return {"value": input_value}
+
+    return {"value": str(input_value)}
```
sonleoracle, line 94: why is this now returning a dict with "value" as the key? does the original `ast.literal_eval(input_str)` do that too?

sonleoracle (approved):

pyagentspec/src/pyagentspec/adapters/langgraph/tracing.py
cesarebernardis: Not really, but the `inputs` extracted from the `kwargs` should do the job instead. This is the last resort among the fallback options, in case the inputs provided in the other parameters do not meet the expectations.

paul-cayet (approved):

dhilloulinoracle:
Merge Gate failed: Build ID #665

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #460

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #666

PR #178, opened 2026-05-06
Improve AgentSpec component loading defaults

ljupche98:


docs/pyagentspec/source/changelog.rst
```diff
@@ -183,6 +187,13 @@ Property titles in Agent Spec must not be empty. This is now enforced by validat
 
 Migration: If your YAML/JSON configurations have properties without titles, you’ll need to set a non-empty, descriptive title for those properties to pass validation. If you generate Agent Spec configurations via the SDK, your code may still work, but we recommend explicitly setting property titles to ensure forward compatibility.
 
+* **MCP stdio transports are blocked by default in Agent Spec loaders**
+
+  Configurations containing ``StdioTransport`` or ``PluginStdioTransport`` will no longer
```
cesarebernardis: `PluginStdioTransport` is not Agent Spec. I guess it comes from wayflow, but it should not be mentioned here

docs/pyagentspec/source/security.rst
```diff
@@ -340,6 +340,22 @@ utilities to avoid unsafe behavior.
...
+Considerations regarding runtime-sensitive components
+-----------------------------------------------------
+
+Some Agent Spec components can influence actions performed by adapters or runtimes
+when a configuration is loaded or run. For example, MCP stdio transports can start
+a local MCP server subprocess on the machine that loads and runs the resulting
+agent.
+
+Only enable runtime-sensitive components from trusted configurations. PyAgentSpec
+loaders block ``StdioTransport`` and ``PluginStdioTransport`` by default. If a
```
cesarebernardis: Same here, no `PluginStdioTransport`

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -8,8 +8,12 @@
 from typing import Any, Dict, List, Optional, Union, cast
 
 from pyagentspec.adapters._agentspecloader import AdapterAgnosticAgentSpecLoader
-from pyagentspec.adapters.langgraph._agentspecconverter import LangGraphToAgentSpecConverter
-from pyagentspec.adapters.langgraph._langgraphconverter import AgentSpecToLangGraphConverter
+from pyagentspec.adapters.langgraph._agentspecconverter import (
+    LangGraphToAgentSpecConverter,
+)
+from pyagentspec.adapters.langgraph._langgraphconverter import (
+    AgentSpecToLangGraphConverter,
```
cesarebernardis: You probably need to run the formatter, these changes look suspicious

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
```diff
@@ -0,0 +1,100 @@
...
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+"""Component allow/block policy helpers used while loading Agent Spec configurations."""
+
+from typing import Iterable, Optional, Set, Type, Union
+
+from pyagentspec.component import Component
+
+ComponentPolicyEntry = Union[str, Type[Component]]
+ComponentPolicyInput = Optional[Union[ComponentPolicyEntry, Iterable[ComponentPolicyEntry]]]
```
cesarebernardis: Nit: I know that most of the times this is used in parameter types, where the `Optional` part is needed, but thinking about the fact that this is a type alias, I think it would make more sense to not have the Optional here, so that the `ComponentPolicyInput` refers to the real type itself, not to the fact that it might not be specified in method parameters (the Optional part of it).

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
```diff
@@ -0,0 +1,100 @@
...
+    if isinstance(field_value, (list, set, tuple)):
+        return [
+            child
+            for inner_field_value in field_value
+            for child in _get_children_direct_from_field_value(inner_field_value)
+        ]
+    return []
+
+
+class ComponentLoadPolicy:
```
cesarebernardis, line 92: I think that this class is not/should not be a public API, I would make it private.

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
```diff
@@ -0,0 +1,100 @@
...
+        component_types = [component_types]
+    elif isinstance(component_types, type) and issubclass(component_types, Component):
+        component_types = [component_types]
+
+    normalized_component_types: Set[str] = set()
+    for component_type in component_types:
+        if isinstance(component_type, str):
+            normalized_component_types.add(component_type)
+        elif isinstance(component_type, type) and issubclass(component_type, Component):
+            normalized_component_types.add(component_type.__name__)
```
cesarebernardis: Components have a field for the name (`component_type`), let's use that.

EDIT: I see that it is not a class property, reason why you did not use it here I guess. Could you try to see if you can make that field a class field?

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def validate_component_type(self, component_type: str) -> None:
+        """Raise if the component type is disallowed by the policy."""
+        if component_type in self.blocked_components:
+            raise ValueError(
+                f"Loading Agent Spec component type `{component_type}` is blocked by policy."
+            )
+        if self.allowed_components is not None and component_type not in self.allowed_components:
+            raise ValueError(
+                f"Loading Agent Spec component type `{component_type}` is not allowed by policy."
```
cesarebernardis: Nit: `blocked` and `not allowed` sound like the same thing, maybe say `is in the block list`, `is not in the allow list`, so that it's clearer why.

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
```diff
@@ -0,0 +1,100 @@
...
+        """Validate a constructed component and all nested child components."""
+        components_to_check = [component]
+        visited_component_ids: set[int] = set()
+        while components_to_check:
+            current_component = components_to_check.pop()
+            if id(current_component) in visited_component_ids:
+                continue
+            visited_component_ids.add(id(current_component))
+
+            self.validate_component_type(current_component.__class__.__name__)
```
cesarebernardis: Let's use `component_type` here too

pyagentspec/src/pyagentspec/adapters/openaiagents/agentspecloader.py
```diff
@@ -46,6 +49,7 @@ def _load_component(
     module_name: Optional module name to stamp in generated code.
     rulepack_version: Optional explicit RulePack version; defaults to SDK version.
     """
+    agent_spec_loader.component_load_policy.validate_component_tree(agentspec_component)
```
cesarebernardis: I am surprised that you need to make it explicit here, but it might be because of the special way this adapter has to treat flows. Can you please double check that you need this, and put a comment explaining why if it's the case

pyagentspec/src/pyagentspec/adapters/_agentspecloader.py
```diff
@@ -8,13 +8,35 @@
...
+from pyagentspec.serialization.componentpolicy import (
+    ComponentLoadPolicy,
+    ComponentPolicyInput,
+)
 
 _RuntimeComponentT: TypeAlias = Any
 _RuntimeRegistryT: TypeAlias = Dict[str, Any]
+_DEFAULT_BLOCKED_COMPONENTS: tuple[str, ...] = (
+    "StdioTransport",
+    "PluginStdioTransport",
```
cesarebernardis: `PluginStdioTransport` is not an Agent Spec thing, so it should not be here. It is in wayflow, right?

I think we need to check two things:
- if it extends the `StdioTransport` class (I guess not, otherwise we should not need to specify it here)
- if it does not, wayflow's adapter should take care of adding this class, as well as all the other potentially dangerous classes it defines in the plugins, in the list of blocked components by default

pyagentspec/src/pyagentspec/serialization/deserializer.py
```diff
@@ -17,25 +17,62 @@
...
 
 class AgentSpecDeserializer:
     """Provides methods to deserialize Agent Spec Components."""
 
-    def __init__(self, plugins: Optional[List[ComponentDeserializationPlugin]] = None) -> None:
+    def __init__(
+        self,
+        plugins: Optional[List[ComponentDeserializationPlugin]] = None,
+        allowed_components: ComponentPolicyInput = None,
+        blocked_components: ComponentPolicyInput = None,
```
cesarebernardis: I think we need to add these parameters also to the `from_*` methods in `Component` (`from_json`, ...).

pyagentspec/src/pyagentspec/serialization/deserializer.py
```diff
@@ -17,25 +17,62 @@
...
+        blocked_components:
+            Optional iterable of component type names or Component classes blocked from loading.
         """
+        component_load_policy = ComponentLoadPolicy(
+            allowed_components=allowed_components,
+            blocked_components=blocked_components,
+        )
         # for early failure when using incorrect plugins
-        _DeserializationContextImpl(plugins=plugins)
+        _DeserializationContextImpl(
```
cesarebernardis: I guess we can use the new `_new_deserialization_context` method

pyagentspec/src/pyagentspec/serialization/deserializer.py
```diff
@@ -17,25 +17,62 @@
...
+            plugins=plugins,
+            allowed_components=component_load_policy.allowed_components,
+            blocked_components=component_load_policy.blocked_components,
+        )
         self.plugins = plugins
+        self.allowed_components = component_load_policy.allowed_components
+        self.blocked_components = component_load_policy.blocked_components
+
+    def _new_deserialization_context(
```
cesarebernardis: Nit: as this is a method, I would call it `_get_new_deserialization_context`.

pyagentspec/tests/test_agentspecloader_stdio.py
```diff
@@ -0,0 +1,76 @@
...
+        _IdentityLoader().load_component(_make_stdio_mcp_tool())
+
+
+def test_adapter_agnostic_loader_respects_allowed_components() -> None:
+    with pytest.raises(ValueError, match="MCPTool.*not allowed by policy"):
+        loader = _IdentityLoader(
+            allowed_components=["StdioTransport"],
+            blocked_components=[],
+        )
+        loader.load_component(_make_stdio_mcp_tool())
```
cesarebernardis, line 115: We need more tests. 

- Play with combinations of blocked and allowed component lists, different types in there (mix of str and Component classes).
- Try to define a new Component class and to add it to block/allow to make sure that it works also with extensions
- Try with hierarchical blocks (a child class of a blocked class should be blocked/allowed)

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
sonleoracle: This overridden `load_component` should be modified too.  `load_component()` calls can pass an already-constructed MCPTool with StdioTransport despite the default blocklist

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #441

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
sonleoracle, line 503: Can we also validate the component before returning here?

```python
 self.component_load_policy.validate_component_tree(component)
 return component, validation_errors
```

sonleoracle, line 521: Should we validate the component types in `new_referenced_components` before storing them? Today a config can include an unused blocked component under `$referenced_components` and still load, because policy validation only happens later if that reference is resolved. For example, an unused `StdioTransport` entry is accepted even with `blocked_components=["StdioTransport"]`, which seems to contradict the docs/changelog wording that configurations containing stdio transports are blocked by default.

sonleoracle (approved):

docs/pyagentspec/source/changelog.rst
ljupche98: Removed it from Agent Spec. But WayFlow has its own AgentSpecLoader and its own Agent Spec plugin components. Even though PyAgentSpec now blocks native StdioTransport by default, WayFlow still needs to pass the same policy through its loader and include its own sensitive components. I created another PR on WayFlow side for this: https://github.com/oracle/wayflow/pull/156

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
ljupche98: It will be needed on WayFlow side, so I kept it public. See [this](https://github.com/oracle/wayflow/pull/156/changes#diff-f9c849140b47d2a9569ecf6136ac29981db39482db4c716cada5e4abe2032276R70) on how it's used in WayFlow

pyagentspec/src/pyagentspec/adapters/openaiagents/agentspecloader.py
ljupche98: For normal components, the adapter uses AdapterAgnosticAgentSpecLoader.load_component, which already validates the component policy. Flows are a special case because they bypass the usual conversion path, so they still need policy check here. I added a comment in the code.

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
cesarebernardis: I see, that's because wayflow does not use the adapter agnostic generalization (while it probably should nowadays). It's ok then.

docs/pyagentspec/source/changelog.rst
cesarebernardis: Yes, that's what we need.

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
ljupche98: I added checks for this case

pyagentspec/src/pyagentspec/adapters/crewai/agentspecloader.py
```diff
@@ -25,9 +26,16 @@ def __init__(
         tool_registry: Optional[Dict[str, Any]] = None,
         plugins: Optional[List[Any]] = None,
         *,
+        allowed_components: Optional[ComponentPolicyInput] = None,
+        blocked_components: Optional[ComponentPolicyInput] = None,
```
dhilloulinoracle, line 39: there is no docstring on this class, and the parent class doesn't appear in the docs.
Would appreciate if you could add a docstring here, explaining in particular the default blocked components, as well as trying to get the parent class to appear in the docs

pyagentspec/src/pyagentspec/serialization/deserializer.py
```diff
@@ -26,16 +27,44 @@
...
         """
         Instantiate an Agent Spec Deserializer.
 
         plugins:
             List of plugins to serialize additional components.
+        allowed_components:
+            Optional iterable of component type names or Component classes allowed to be loaded.
+            If omitted, all component types are allowed unless blocked.
+        blocked_components:
+            Optional iterable of component type names or Component classes blocked from loading.
```
dhilloulinoracle, line 55: please explain what are the components blocked by default and why

pyagentspec/src/pyagentspec/component.py
```diff
@@ -1092,6 +1098,11 @@ def from_yaml(
             main component.
         plugins:
             List of plugins to deserialize additional components.
+        allowed_components:
+            Optional iterable of component type names or Component classes allowed to be loaded.
+            If omitted, all component types are allowed unless blocked.
+        blocked_components:
+            Optional iterable of component type names or Component classes blocked from loading.
```
dhilloulinoracle, line 1105: put the components blocked by default and why

pyagentspec/src/pyagentspec/component.py
```diff
@@ -1151,6 +1172,11 @@ def from_json(
...
```
dhilloulinoracle, line 1186: same

pyagentspec/src/pyagentspec/component.py
```diff
@@ -1210,6 +1246,11 @@ def from_dict(
...
```
dhilloulinoracle, line 1267: same

dhilloulinoracle:
Looks like you are considering the class hierarchy.
What happens in those cases:
- allow LLM (parent class), block OpenAI?
- block LLM (parent class), allow OpenAI?

I think ideally it should find the most precise rule (sub class) and use that.

Behavior in those cases should be documented in the different docstrings.
We should test the chosen behavior

ljupche98:
> Looks like you are considering the class hierarchy. What happens in those cases:
> 
> * allow LLM (parent class), block OpenAI?
> * block LLM (parent class), allow OpenAI?
> 
> I think ideally it should find the most precise rule (sub class) and use that.
> 
> Behavior in those cases should be documented in the different docstrings. We should test the chosen behavior

@dhilloulinoracle I updated the logic to consider the most precise rule. With the latest changes:
- allow LLM (parent class), block OpenAI: OpenAI is blocked
- block LLM (parent class), allow OpenAI: OpenAI is allowed

pyagentspec/src/pyagentspec/serialization/componentpolicy.py
ljupche98: I checked this, but making `component_type` a class-level field is not small and local change. Making it a reliable class attribute would require a broader change to the component model and registration/deserialization flow

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
```diff
@@ -38,6 +39,16 @@ class AgentSpecLoader(AdapterAgnosticAgentSpecLoader):
...
         Optional ``RunnableConfig`` to pass to created runnables/graphs.
+    allowed_components:
+        Optional iterable of Agent Spec component type names or Component classes allowed
+        to be loaded. If omitted, all component types are allowed unless blocked.
+    blocked_components:
+        Optional iterable of Agent Spec component type names or Component classes blocked
+        from loading. If omitted, stdio MCP transports are blocked by default.
+        Resolvable type names and Component classes also match subclasses; unresolved
+        type names match only the exact serialized component type. When allow and
+        block entries both match, the most specific entry wins; block entries win
```
cesarebernardis: Can you be a little more precise and explain what the "most specific entry" wins? I guess you mean the closest class match in the class hierarchy?

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
```diff
@@ -487,8 +497,91 @@ def _load_component_with_plugin(
...
 
         component.min_agentspec_version = agentspec_version
+        self.component_load_policy.validate_component_tree(component)
         return component, validation_errors
 
+    def _validate_component_config_types(
+        self,
+        content: object,
+        annotation: Optional[type] = None,
+    ) -> None:
```
cesarebernardis: Looks like a duplicate of something that should be already happening while building components, is there a reason why we cannot use the same logic and just inject validation?

dhilloulinoracle (approved):

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
ljupche98: Yes, this overlaps with the recursive traversal done during normal component construction. The difference is that $referenced_components are loaded lazily. If a referenced component is never resolved, it is never built and the other validation never checks it.

pyagentspec/src/pyagentspec/adapters/langgraph/agentspecloader.py
ljupche98: Yes, it's the closest ancestor class in the class hierarchy. I updated the relevant docstrings

pyagentspec/src/pyagentspec/serialization/deserializationcontext.py
cesarebernardis: But if a referenced component is never resolved, it means that it is never built, and therefore never used in the component being deserialized. Did I get it right? Because if that's the case, I think it's not a problem.

ljupche98: Agreed to remove it. If a blocked component is actually used, it'll be caught by the other check

cesarebernardis (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #467

dhilloulinoracle:
Merge Gate failed: Build ID #675

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #676

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #677

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #678

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #471

dhilloulinoracle:
Merge Gate failed: Build ID #679

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #472

dhilloulinoracle:
Merge Gate failed: Build ID #680

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #681

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #683

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #684

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #685

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #686

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Merge Gate failed: Build ID #687

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #479

dhilloulinoracle:
Merge Gate failed: Build ID #688

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #480

dhilloulinoracle:
Merge Gate failed: Build ID #689

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #481

dhilloulinoracle:
Merge Gate failed: Build ID #690

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #482

dhilloulinoracle:
Merge Gate failed: Build ID #691

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #483

dhilloulinoracle:
Merge Gate failed: Build ID #692

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #484

dhilloulinoracle:
Merge Gate failed: Build ID #693

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #485

dhilloulinoracle:
Merge Gate failed: Build ID #694

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #486

dhilloulinoracle:
Merge Gate failed: Build ID #695

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #487

dhilloulinoracle:
Merge Gate failed: Build ID #696

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #488

dhilloulinoracle:
Merge Gate failed: Build ID #697

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #489

dhilloulinoracle:
Merge Gate failed: Build ID #698

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #490

dhilloulinoracle:
Merge Gate failed: Build ID #699

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #491

dhilloulinoracle:
Merge Gate failed: Build ID #700

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #492

dhilloulinoracle:
Merge Gate failed: Build ID #701

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #493

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #702

PR #190, opened 2026-05-14
Add includeSensitiveFields opt-in to serializers (TS + Python)

spichen:
## Summary

- Adds `includeSensitiveFields` option to `AgentSpecSerializer.toJson()`, `toYaml()` (TS) and `to_json()`, `to_yaml()`, `to_dict()` (Python)
- When set to `true`/`True`, sensitive fields (API keys, TLS certs, headers) are serialized inline instead of being redacted
- Default behaviour is unchanged — sensitive fields are still stripped/referenced unless the consumer explicitly opts in

## Changes

**TypeScript**
- `SerializationContext` constructor refactored from 6 positional args to an options object (eliminates `undefined, undefined` placeholders at call sites)
- `isFieldSensitive()` short-circuits when `includeSensitiveFields` is set
- `toJson` / `toYaml` option types updated

**Python**
- `SerializationContext` base class gains `is_field_sensitive(field_info)` method that encapsulates the flag — plugins call this instead of reading the raw flag and inverting it
- `include_sensitive_fields: bool = False` default added to the abstract base (prevents silent `AttributeError` for custom subclasses)
- `to_dict` / `to_json` / `to_yaml` overloads and implementations updated

### DIFF
--- a/pyagentspec/src/pyagentspec/serialization/serializer.py
+++ b/pyagentspec/src/pyagentspec/serialization/serializer.py
@@ -58,6 +58,8 @@ class AgentSpecSerializer:
     def to_yaml(
         self,
         component: Component,
+        *,
+        include_sensitive_fields: bool = False,
     ) -> str: ...
 
     @overload
@@ -65,6 +67,8 @@ class AgentSpecSerializer:
         self,
         component: Component,
         agentspec_version: Optional[AgentSpecVersionEnum] = None,
+        *,
+        include_sensitive_fields: bool = False,
     ) -> str: ...
 
     @overload
@@ -386,6 +390,8 @@ class AgentSpecSerializer:
     def to_dict(
         self,
         component: Component,
+        *,
+        include_sensitive_fields: bool = False,
     ) -> ComponentAsDictT: ...
 
     @overload
@@ -393,6 +399,8 @@ class AgentSpecSerializer:
         self,
         component: Component,
         agentspec_version: Optional[AgentSpecVersionEnum],
+        *,
+        include_sensitive_fields: bool = False,
     ) -> ComponentAsDictT: ...
 
     @overload
--- a/pyagentspec/src/pyagentspec/serialization/serializer.py
+++ b/pyagentspec/src/pyagentspec/serialization/serializer.py
@@ -166,13 +166,16 @@ class AgentSpecSerializer:
         export_disaggregated_components:
             Whether to export the disaggregated components or not. Defaults to ``False``.
         include_sensitive_fields:
-            If ``True``, sensitive fields like API keys and certificate paths are written to the
-            output as plain text rather than replaced with ``$component_ref`` placeholders.
-            Defaults to ``False``.
+            If ``False`` (default), non-empty sensitive fields are exported as
+            ``$component_ref`` placeholders. If ``True``, their values are included in the
+            returned serialization. Use this only for trusted local workflows; the returned
+            value may contain secrets or other sensitive data and should not be logged,
+            committed, or shared.
 
             .. warning::
 
-                The output will contain credentials in plain text. Treat it accordingly.
+                Enabling this option can expose API keys, credentials, file paths, headers, DSNs,
+                or other sensitive values in the returned data.
 
         Returns
         -------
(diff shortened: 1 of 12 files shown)


dhilloulinoracle:
Hey @spichen ,

Thanks for this new contribution!

We will review it internally a bit more than other things as it has security implications, but I like the idea of it as it would simplify things quite a bit.

pyagentspec/src/pyagentspec/serialization/serializer.py
cesarebernardis, line 502: Relevant docstrings need to be updated with the new parameter

cesarebernardis, line 578: I would raise a warning if `include_sensitive_fields` is True to inform the user that the serialized value returned might contain sensitive information, and that it should not be shared (plus a corresponding test)

dhilloulinoracle: Warning once if `include_sensitive_fields` is set makes sense, warning also when a sensitive field was actually serialized would also be great so that people can know what ended being serialized and might be good to pay attention to

dhilloulinoracle:
Internal regression failed: Build ID #484

dhilloulinoracle:
Internal regression failed: Build ID #490

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #491

cesarebernardis (approved):

dhilloulinoracle:
Hi @spichen ,
@cesarebernardis pushed a few changes to get the CI green and he added the changelog.

If there is nothing else you want to add here, we can merge it :)

dhilloulinoracle (approved):

spichen:
> Hi @spichen , @cesarebernardis pushed a few changes to get the CI green and he added the changelog.
> 
> If there is nothing else you want to add here, we can merge it :)

@dhilloulinoracle nothing else from my side. We can merge. Thanks!

pyagentspec/src/pyagentspec/serialization/serializer.py
vvenz1n, line 75: For completeness `to_yaml` and `to_dict` should have overloads covering the following calls as well respectively

* `AgentSpecSerializer().to_yaml(component, include_sensitive_fields=True)`
* `AgentSpecSerializer().to_dict(component, include_sensitive_fields=True)`

vvenz1n (changes requested):
We should update documentation to reflect this new feature.

**1) docs/pyagentspec/source/security.rst around line 70**

"Never embed API keys, passwords, or other secrets directly in Agent Spec."
->
"Do not embed API keys, passwords, or other secrets directly in Agent Spec configurations intended for storage, sharing, version control, logging, or deployment."

**2) docs/pyagentspec/source/security.rst around line 111**

Replace opening guidance, until before "Furthermore. ..." with

```
.. important::

  Do not embed API keys, passwords, or other secrets directly in Agent Spec
  configurations intended for storage, sharing, version control, logging, or deployment.

Agent Spec is a declarative language for agents, tools, and flows.
It may be stored, versioned, or inspected; it is **not a secure medium for sensitive data**.
Hardcoding secrets in shared or persistent representations increases the risk of accidental
exposure and must be avoided.

Instead of embedding secrets in the representation, ensure that secrets are injected securely
at runtime via:

* Integration with a secrets manager (e.g., OCI Vault)
* Use instance principal when working on OCI Computes
* Use Kubernetes/Docker secrets if the deployments supports it

If sensitive values are explicitly included in an Agent Spec configuration for a trusted
local workflow, treat that configuration as secret material. Do not commit, log, share,
or persist it except in systems approved for storing secrets.
```

**3) docs/pyagentspec/source/agentspec/language_spec_nightly.rst near line disaggregated configuration section, around line 2860**

"Agent Spec exported configurations should never contain sensitive information."
--> 
"Agent Spec configurations intended for storage, sharing, version control, logging,
or deployment should not contain sensitive information. Sensitive values should be
represented as references and supplied through trusted runtime or deserialization
mechanisms."

**4) docs/pyagentspec/source/agentspec/language_spec_nightly.rst around line 2944**

"- **When exporting configurations**: Any sensitive field must be excluded from the configurations."
-->
"- **When exporting configurations**: Sensitive fields should be excluded by default and represented as references. Any export that intentionally includes sensitive values must be treated as secret material, not as an ordinary portable Agent Spec configuration."

vvenz1n (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #505

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #717

PR #192, opened 2026-05-20
Spec clarifications

vvenz1n:
1. clarify branching support in nightly spec
2. note node table horizontal scrolling
3. clarify duplicate inline definitions
4. clarify component reference scope
5. exclude null from string conversion rule
6. clarify default value compatibility
7. clarify reference validity rules
8. clarify node parameter constraints
9. clarify map node IO rules

### DIFF
--- a/docs/pyagentspec/source/agentspec/language_spec_nightly.rst
+++ b/docs/pyagentspec/source/agentspec/language_spec_nightly.rst
@@ -189,6 +189,9 @@ In order to further simplify input-output connections, we define some simple typ
   The convention for the conversion is the one adopted by most programming languages:
   0 refers to ``false``, while any other number refers to ``true``.
 
+Default values follow the same compatibility rules. If a property defines ``default``,
+that value must be compatible with the property's declared schema.
+
 These rules apply recursively in complex types:
 
 - To the types of the elements of an ``array``;
--- a/docs/pyagentspec/source/agentspec/language_spec_nightly.rst
+++ b/docs/pyagentspec/source/agentspec/language_spec_nightly.rst
@@ -82,6 +82,10 @@ Note that a subcomponent reused in multiple parts of a complex nested component
 a flow) must be defined using a component reference. If two components in an
 Agent Spec configuration use the same `id`, the configuration will be considered
 invalid.
+Duplicate inline definitions are invalid even when they appear in different
+fields. For example, a node with the same ``id`` must not be defined inline in
+more than one of ``start_node``, ``nodes``, or ``$referenced_components``;
+define it once and use ``$component_ref`` everywhere else.
 
 
 Input/output schemas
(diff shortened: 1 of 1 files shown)


docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -171,13 +193,16 @@ but an output of type ``string`` cannot connect to an input of type ``number``.
 
 In order to further simplify input-output connections, we define some simple type compatibility rules that are accepted by Agent Spec.
 
-- Every type can be converted to string.
+- Every type except ``null`` can be converted to string.
```
cesarebernardis, line 196: This is a new limitation and a breaking change, we kept the conversion of any type to string for simplicity. I don't think we should add this.

vvenz1n: json-schema says that null can have exactly one value, and it’s null

https://json-schema.org/understanding-json-schema/reference/null#null

Given that strictness, I think it’s better to not allow converting it.

E.g. anyOf{string, null}  -> string , should not work IMO.

The target here wants a string, it doesn’t expect (or know) what to do with null

“null”  means explicitly “no value” or “absence of value”, so converting it into a value seems not correct to me.

oracle-nish: We might have several instances in an executor implementation where we would need anyOf{string, null} -> string.

1st example that comes to mind is:
An AgentSpec has a ToolNode / LLMNode accepts an output property which is an anyOf{string, null} with a default value of "null". Here, we may have an outgoing dataflow connection to an BranchingNode which needs to make a control flow decision based on the null value. Since we currently allow only string representations of mappings, executors might need to allow the anyOf{string, null} -> string conversion to happen to do mapping string compare check. 

2nd (rather simpler) example: We may simply have a requirement to display a stringified "null" value to the user for specific cases where either a ToolNode/LLMNode/AgentNode returns null, which would require the data flow connection from anyOf{string, null} to string to be valid. 

Hence, I do agree with @cesarebernardis that this rule can be made more lenient.

docs/pyagentspec/source/agentspec/language_spec_nightly.rst
```diff
@@ -1798,25 +1829,30 @@ A more detailed description of each node follows.
               - null, each output is aggregated through value concatenation (append)
 
       - Inferred from the inner structure (as defined in FlowNode).
-        The names of the inputs will be the ones of the inner flow,
-        complemented with the ``iterated_`` prefix. Their type is
+        The inputs must exactly match the inputs of the inner flow,
+        with each name complemented with the ``iterated_`` prefix.
+        Extra or missing inputs are invalid. Their type is
         ``Union[inner_type, List[inner_type]]``, where ``inner_type``
         is the type of the respective input in the inner flow.
```
oracle-nish, line 1836: @vvenz1n Specific to MapNode, we agreed that the subflow's input/output types need not be exactly the same as the MapNode's "inner_type" within the input/output types, and that they can be castable. Can we mention this explicitly here?

PR #199, opened 2026-06-09
fix(agent-framework): pin supported beta install

sonleoracle:
## Summary
- Pin the agent-framework extra to the known-compatible beta package
- Rely on that pinned agent-framework package to resolve the matching agent-framework-core dependency
- Leave constraints files unchanged
- Disable OpenAI Agents SDK trace export in adapter tests so the fake test API key does not trigger background posts to api.openai.com during CI

## Validation
- Fresh temp install of pyagentspec[agent-framework] without an explicit agent-framework-core requirement resolved agent-framework==1.0.0b260130 and agent-framework-core==1.0.0b260130
- pip check passed in that temp install
- Pre-commit hooks passed via shared agent-spec venv
- Targeted OpenAI Agents live/simple-flow plus security/datastore regression passed locally with the shared venv and LLM env

### DIFF
--- a/pyagentspec/tests/adapters/openaiagents/conftest.py
+++ b/pyagentspec/tests/adapters/openaiagents/conftest.py
@@ -14,16 +14,11 @@ like `pyagentspec.adapters.openaiagents` and `agents` resolve during tests.
 import os
 import sys
 from pathlib import Path
-from typing import Any
+from typing import Any, Iterator
 
 import pytest
 
 from ..conftest import skip_tests_if_dependency_not_installed
-
-# OpenAI Agents exports traces to api.openai.com when tracing is enabled and an
-# OPENAI_API_KEY is present. Adapter tests use a fake key, so disable SDK trace
-# export before test modules import `agents`.
-os.environ["OPENAI_AGENTS_DISABLE_TRACING"] = "1"
 
 
 def _add_path(p: Path) -> None:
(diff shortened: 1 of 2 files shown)


dhilloulinoracle:
Internal regression failed: Build ID #498

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #499

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #500

pyagentspec/tests/adapters/openaiagents/conftest.py
```diff
@@ -11,12 +11,20 @@
...
 from typing import Any
 
+import pytest
+
 from ..conftest import skip_tests_if_dependency_not_installed
 
+# OpenAI Agents exports traces to api.openai.com when tracing is enabled and an
+# OPENAI_API_KEY is present. Adapter tests use a fake key, so disable SDK trace
+# export before test modules import `agents`.
+os.environ["OPENAI_AGENTS_DISABLE_TRACING"] = "1"
```
cesarebernardis: I think it's better to set and reset this as part of the `_disable_openai_agents_trace_export` fixture (that could yield at this point) so that we restore the old value and we don't pollute developer's environment.

Check for example `_disable_openai_api_key`

cesarebernardis (approved):

paul-cayet (approved):

pyagentspec/tests/adapters/openaiagents/conftest.py
sonleoracle: done

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #501

cesarebernardis:
Not applicable anymore, agent-framework was upgraded

PR #207, opened 2026-06-29
Add TypeScript Mastra adapter

Devx228:
Adds initial Mastra adapter support for the TypeScript SDK.
This includes Agent Spec -> Mastra loading, Mastra-style config -> Agent Spec export, model/tool mapping, datastore target mapping for in-memory and Postgres storage, initial linear Flow -> Mastra Workflow conversion, docs, tests, and adapter examples.  

Credit: this work was developed with **Darsh Patel** (@darshpatel1052). The branch is squashed under my OCA verified contributor account for contribution processing.

### DIFF
--- a/tsagentspec/docs/mastra/README.md
+++ b/tsagentspec/docs/mastra/README.md
@@ -1,0 +1,314 @@
+# Mastra Adapter
+
+The Mastra adapter connects Agent Spec's portable agent definitions with Mastra
+agent, tool, storage, and workflow runtime objects. It supports both directions:
+
+- load Agent Spec components into a Mastra runtime
+- export serializable Mastra-style agent configs back into Agent Spec YAML or JSON
+
+The adapter keeps Mastra optional at package install time. `agentspec` does not
+make `@mastra/core`, `@mastra/pg`, or other Mastra packages hard dependencies.
+When the adapter is used, the default runtime binder resolves Mastra packages
+from the consuming application.
+
+## Design
+
+Agent Spec is the portable contract. Mastra is the runtime that executes it.
+The adapter keeps those responsibilities separate:
+
+- Agent Spec defines the agent, model config, prompt, tools, schemas, flows, and
+  datastore configuration.
+- The adapter supplies default Mastra `Agent`, `createTool`, `createWorkflow`,
+  and `createStep` bindings for normal loading.
+- The application supplies executable behavior such as tool functions, workflow
+  node executors, storage provider packages, or custom runtime overrides.
+- Resolver hooks translate framework-specific model and tool details without
+  forcing those choices into the Agent Spec core model.
+
+This follows the same adapter shape as the Python integrations: a loader
+converts Agent Spec into a framework runtime, while an exporter converts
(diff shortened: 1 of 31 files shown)


tsagentspec/docs/mastra/README.md
```diff
@@ -0,0 +1,314 @@
...
+`convertAgentSpecFlowToMastraWorkflow` converts simple linear Agent Spec flows
+into Mastra workflow runtime configs.
+
+Supported linear shape:
+
+```text
+StartNode -> LlmNode/ToolNode/AgentNode... -> EndNode
+```
+
+The converter rejects branching, cycles, disconnected nodes, unsupported node
```
cesarebernardis, line 241: I would mention that also data flow edges are not supported, if I am not mistaken I did not see their translation either

tsagentspec/src/adapters/mastra/datastores.ts
```diff
@@ -0,0 +1,103 @@
...
+): MastraPostgresDatastoreTarget => {
+  const id = options.id ?? datastore.id;
+  const connection = datastore.connectionConfig;
+  // Datastore resolvers only build an import/config target. They should never
+  // open sockets, run migrations, or require the Mastra provider package.
+  const config: MastraPostgresStoreConfig = {
+    id,
+    connectionString: connection.url,
+    user: connection.user,
+    password: connection.password,
```
cesarebernardis, line 65: It looks like user and password are ignored by mastra, as they are extracted from the URL. Can you check if that's true? I think you either have to:
- construct an authenticated URL, e.g. `postgresql://user:password@localhost:5432/agentspec` (with correct URL escaping), or
- use Mastra’s host-based configuration form, where username/password are supported fields.

tsagentspec/src/adapters/mastra/mastra-converter.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      const toolConfig = this.toolResolver(tool, {
+        agent,
+        toolRegistry: this.toolRegistry,
+      });
+      tools[tool.name] = this.runtime.createTool(toolConfig);
+    }
+
+    return this.runtime.createAgent({
```
cesarebernardis, line 52: Nit: could you reject explicitly configurations that are not supported (e.g., human in the loop)? It would be good to see an explicit message saying that it's not supported, here and in the other conversions that are partially supported.

tsagentspec/src/adapters/mastra/agentspec-converter.ts
```diff
@@ -0,0 +1,620 @@
...
+  nativeToolExporter?: (
+    tool: unknown,
+    context: MastraToolExportContext,
+  ) => Tool | undefined;
+  defaultModelName?: string;
+  nativeModelBaseUrls?: Record<string, string | undefined>;
+  includeModelApiKey?: boolean;
+};
+
+export class MastraToAgentSpecConverter {
```
cesarebernardis, line 49: I am not sure about how exactly docstrings work in typescript, but probably worth having minimal ones in public classes and their public methods.

PR #210, opened 2026-07-25
feat(adapters/langgraph): make agent structured output not hang or  silently vanish

spichen:
## Problem

`create_agent(response_format=...)` ties the agent's exit condition to getting a parseable structured response. The adapter set it for any agent with declared outputs, so a model that won't produce one broke two ways.

With no tools the agent loops. `create_agent` sets `recursion_limit=9999` where LangGraph's default is 25, so it burns thousands of model calls and then raises a `GraphRecursionError` naming neither the agent nor the schema.

With tools it's quieter and worse. Routing treats a turn with no tool calls as done, so the agent exits, `structured_response` is never set, and a two-output agent returns zero outputs with no error ([langchain#36349](https://github.com/langchain-ai/langchain/issues/36349), open).

Nothing upstream bounds this. `ToolStrategy.handle_errors` only covers malformed tool calls; `True`, `False` and a custom message all still hang. See also [langgraph#6731](https://github.com/langchain-ai/langgraph/issues/6731) (closed, not planned) and this [forum thread](https://forum.langchain.com/t/create-agent-toolstrategy-tools-structured-output-not-working-across-models-langchain-1-0-2/2293). `setup.py` allows `langchain>=1.2.0` unbounded, so it isn't one bad version.

## Fix

An agent whose only declared output is a string needs no structured generation, so `response_format` is skipped and the value read from the final message. `LlmNodeExecutor` already worked this way; `AgentNode` now matches it.

Everything else keeps `response_format` plus a `StructuredOutputGuard` middleware. `after_model` bounds the loop by counting consecutive turns that end in an `AIMessage` with no tool calls. `after_agent` catches the silent exit, which that budget never sees. Turns with tool calls don't count, since under `ToolStrategy` the structured response is itself a tool call.

`is_single_string_output` moved to `adapters/_utils.py`, replacing two copies of the predicate in `LlmNodeExecutor`, one of them dead. `AgentMiddleware` now comes from `_types.py` as a `LazyType`, so the guard is a plain module-level class and the module still imports without the `langgraph` extra.

## Testing

Against a model that ignores structured output:

| | before | after |
|---|---|---|
| single string output | `GraphRecursionError` (9999 calls) | `{'answer': '42'}` |
| multi-field, no tools | `GraphRecursionError` (9999 calls) | error after 3 calls |
| multi-field, with a tool | silently returned 0 of 2 outputs | error |
| multi-field, compliant model | works | works |

That last row is the one that matters: neither hook may fire on a model that does produce the response.

454 passing across `tests/adapters`, `tests/serialization` and `tests/validation` under `SKIP_LLM_TESTS=1`, pinned to `constraints/constraints.txt`; failures and skips unchanged from `main`. `black`, `isort`, `pyflakes`, `mypy` clean. The 10 new tests build no LLM config, so they run offline instead of skipping.

### DIFF
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/_agent_output_guard.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/_agent_output_guard.py
@@ -21,19 +21,23 @@ string avoids the problem entirely, since ``response_format`` is skipped for it 
 :func:`pyagentspec.adapters._utils.is_single_string_output`).
 """
 
-from typing import Any, Dict, List, NoReturn, Optional
+from functools import lru_cache
+from typing import Annotated, Any, Dict, List, NoReturn, Optional
 
 from typing_extensions import NotRequired
 
-from pyagentspec.adapters.langgraph._types import AgentMiddleware, AgentState
+from pyagentspec.adapters.langgraph._types import (
+    AgentMiddleware,
+    AgentState,
+    langchain_middleware_types,
+)
 
 # A compliant run spends at most one turn without a structured response, so a small
 # bound separates "model is working on it" from "model will never comply".
 DEFAULT_MAX_STRUCTURED_OUTPUT_ATTEMPTS = 3
 
-# Must match the field on _StructuredOutputAttemptState. Dunder-wrapped so it cannot
-# collide with a field the agent's own state schema declares.
-_ATTEMPTS_STATE_KEY = "__structured_output_attempts__"
+# Must match the private field on _StructuredOutputAttemptState.
+_ATTEMPTS_STATE_KEY = "_pyagentspec_structured_output_attempts"
 
 
 class StructuredOutputNotProducedError(RuntimeError):
@@ -43,10 +47,19 @@ class StructuredOutputNotProducedError(RuntimeError):
     """
 
 
-class _StructuredOutputAttemptState(AgentState):
-    # On the state rather than the instance, so the count is per run and not shared
-    # between concurrent ones.
-    __structured_output_attempts__: NotRequired[int]
+@lru_cache
+def _structured_output_attempt_state_schema() -> type:
+    """Create the LangChain state extension when its optional dependency is available."""
+
+    class _StructuredOutputAttemptState(AgentState):
+        # On the state rather than the instance, so the count is per run and not shared
+        # between concurrent ones. LangChain omits PrivateStateAttr fields from user input
+        # and output schemas.
+        _pyagentspec_structured_output_attempts: NotRequired[
+            Annotated[int, langchain_middleware_types.PrivateStateAttr]
+        ]
+
+    return _StructuredOutputAttemptState
 
 
 class StructuredOutputGuard(AgentMiddleware):
@@ -55,8 +68,6 @@ class StructuredOutputGuard(AgentMiddleware):
     Install only when ``response_format`` is set. Without it, a turn with no tool calls
     just means the agent is done and an empty ``structured_response`` is fine.
     """
-
-    state_schema = _StructuredOutputAttemptState
 
     def __init__(
         self,
@@ -67,6 +78,7 @@ class StructuredOutputGuard(AgentMiddleware):
         max_attempts: int = DEFAULT_MAX_STRUCTURED_OUTPUT_ATTEMPTS,
     ) -> None:
         super().__init__()
+        self.state_schema = _structured_output_attempt_state_schema()
         self.agent_name = agent_name
         self.output_titles = output_titles
         self.model_id = model_id
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/_agent_output_guard.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/_agent_output_guard.py
@@ -31,8 +31,9 @@ from pyagentspec.adapters.langgraph._types import AgentMiddleware, AgentState
 # bound separates "model is working on it" from "model will never comply".
 DEFAULT_MAX_STRUCTURED_OUTPUT_ATTEMPTS = 3
 
-# Must match the field on _StructuredOutputAttemptState.
-_ATTEMPTS_STATE_KEY = "structured_output_attempts"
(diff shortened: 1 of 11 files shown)


pyagentspec/src/pyagentspec/adapters/langgraph/_structured_output.py
```diff
@@ -0,0 +1,103 @@
...
+from typing_extensions import NotRequired
+
+from pyagentspec.adapters.langgraph._types import AgentMiddleware, AgentState
+
+# A compliant run spends at most one turn without a structured response, so a small
+# bound separates "model is working on it" from "model will never comply".
+DEFAULT_MAX_STRUCTURED_OUTPUT_ATTEMPTS = 3
+
+# Must match the field on _StructuredOutputAttemptState.
+_ATTEMPTS_STATE_KEY = "structured_output_attempts"
```
cesarebernardis: Probably best to use a less common key to use in the shared state (e.g., surround it with double underscore `__`)

pyagentspec/src/pyagentspec/adapters/langgraph/_agent_output_guard.py
```diff
@@ -0,0 +1,103 @@
...
+            f"output to get the model's free text instead, or use a model that supports "
+            f"structured output."
+        )
+
+    # No async variants: LangGraph routes async runs to the sync hooks, and both are pure.
+    def after_model(self, state: Any, runtime: Any) -> Optional[Dict[str, int]]:
+        messages = state.get("messages") or []
+        last = messages[-1] if messages else None
+        if last is None or getattr(last, "type", None) != "ai":
+            return None
```
cesarebernardis, line 114: Should we reset the count instead of returning None?

Maybe worth depicting the cases when we might enter this method to know better what to do in each case, and write a comment about it that includes our assumption and our decision on what to do.

pyagentspec/tests/adapters/langgraph/test_agent_output_guard.py
```diff
@@ -0,0 +1,202 @@
+# Copyright © 2025, 2026 Oracle and/or its affiliates.
+#
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
```
cesarebernardis, line 5: Nit: this file is testing more than just "structured_output", as there are some tests that are specific to agents and their new guard. Maybe worth giving this file another more specific name, same for the name of the main file this one is testing.

cesarebernardis (approved):
Thank you for the contribution, looks good to me. I just left a couple of minor comments, let me know when you have addressed them. I would also recommend to rebase on updated main. I will take over from there and merge the PR.

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #567

fede-kamel (commented):
Checked out the PR head (24610c2) and ran `tests/adapters/langgraph`, `tests/adapters/test_utils.py` and the template-rendering tests with `SKIP_LLM_TESTS=1`: 160 passed, 90 skipped. Then reproduced the claims with a `FakeMessagesListChatModel` injected through the adapter (patched `_llm_convert_to_langgraph`, `recursion_limit=12`), both as a standalone agent and as an `AgentNode` inside a Flow:

| case | result on this PR |
|---|---|
| 2 outputs, no tools, model answers in prose | `StructuredOutputNotProducedError` after 3 model calls |
| 2 outputs, one tool, tool call then prose (standalone and in an `AgentNode`) | `StructuredOutputNotProducedError` after 2 calls ("ended its run without one") |
| 2 outputs, one tool, compliant structured call | terminates, outputs correct, guard silent |
| single string output without default, prose (`AgentNode`) | `{'llm_output': '180'}` |
| single string output **with** `default: ''`, prose (`AgentNode`) | `{'llm_output': ''}` (finding 1) |
| 2 outputs, one tool, structured call with invalid arguments repeated | `GraphRecursionError` after 6 model calls at limit 12 (finding 3) |

So the guard behaves as described, including inside Flows. Three findings and one housekeeping note:

1. **`_node_execution.py`, `extract_outputs_from_invoke_result`, new single-string branch.** Declared defaults are merged into `outputs` before the `if title not in outputs` check, so a single string output that has a default returns the default instead of the final message. That is the shape of the CTS `simple_agent_component.yaml` (`llm_output` with `default: ''`), which is one of the configs behind #227, so the CTS agent tests would still get `''`. Deciding "already provided" from the result rather than from `outputs` fixes it:

   ```python
   provided = title in result or title in dict(result.get("structured_response") or {})
   if not provided:
       ...  # read the final message
   ```

2. **Same branch, `content` type.** `messages[-1].content` can be a list of content blocks (`AIMessage(content=[{"type": "text", "text": "180"}])`); the node executor then `json.dumps` it into `'[{"type": "text", "text": "180"}]'`. `messages[-1].text` (a property in langchain-core 1.x) returns the concatenated text.

3. **Counter reset on any tool call.** A model that keeps calling the structured-output tool with invalid arguments never increments the counter (each turn has a tool call), and `ToolStrategy(handle_errors=True)` keeps retrying, so that case is still bounded only by the recursion limit (9999 under `create_agent`, i.e. thousands of calls). Is that in scope here? Two cheap options would be counting consecutive turns whose only tool call is the structured-output tool, or documenting that `handle_errors` governs it.

4. The branch is based on 6f0b6ae; `main` has since reorganized the changelog (1362da5), so the changelog hunk will conflict on rebase.

Two notes on interplay with PRs of mine, for the maintainers' ordering decision:

- #260 (for #224) also edits `extract_outputs_from_invoke_result`: when no structured response exists it recovers the outputs from the final message if they fit the declared schema, and otherwise falls back to defaults with a warning. With this PR the prose-after-tool case raises instead, so the two choose different behaviour for the same situation. If this PR lands first I will cut #260 down to the default-precedence fix from finding 1; alternatively the guard could attempt that recovery before failing.
- #261 passes the flow run's `recursion_limit` on to `AgentNode` agents; it is independent of this PR and would also bound finding 3 inside Flows.

PR #211, opened 2026-07-25
feat(adapters/langgraph): compile ManagerWorkers into a hierarchical graph

spichen:
`ManagerWorkers` had no LangGraph support: converting one fell through `_convert` and raised. This compiles it into a hierarchical graph.

The manager becomes a react agent holding one synthetic `delegate_to_<worker>` tool per worker. The parent graph intercepts that tool call in a conditional edge and fans out one `Send` per delegation, so a single manager turn can delegate to several workers at once. Each worker runs in an isolated message context and replies with a `ToolMessage` matched to its `tool_call_id`. Every call gets answered independently, since an unanswered one leaves a dangling call in the manager's next turn.

Workers are converted recursively into their own `CompiledStateGraph` and added as subgraph nodes, so `astream_events` still shows the parent/child boundary (`subgraph=True`) for tracing and SSE streaming. Nested `ManagerWorkers` compose through `convert` with no extra work.

`ManagerWorkers` now infers its inputs and outputs from the group manager, so it can sit inside a flow: an `AgentNode` wrapping one exposes input ports, and a `DataFlowEdge` into it resolves. The graph runs over `MessagesState`, which can't carry structured inputs inward, so node inputs are rendered into the group-manager prompt and the satisfied ports are dropped. `AgentNodeExecutor` compiles and caches that graph by rendered prompt, the same key it already uses for plain react agents.

The delegation protocol is deliberately visible: `delegate_to_<worker>` calls and their lifecycle events stream like any other tool call, because "the manager handed task X to worker Y" is usually the most informative thing a run produces. Nothing wraps `astream_events` to scrub it. A consumer that would rather not render the routing plumbing can recognize it with `is_delegation_tool_name`.

## Structure

Helpers live in a new `_managerworkers.py` rather than growing `_langgraphconverter.py`; `_manager_workers_convert_to_langgraph` stays on the converter next to its siblings, and the flow-step compile sits on `AgentNodeExecutor` next to its react-agent equivalent.

Wrapping `stream`/`astream` in an execution span was already duplicated between the Agent and Flow paths, and ManagerWorkers needed a third copy. That wrapper is now a single `patch_with_execution_span` in `_execution_span.py`, parameterized by span and event factories, and all three paths call it.


pyagentspec/tests/adapters/conftest.py
```diff
@@ -112,6 +114,34 @@ def _skip(*_args, **_kwargs):
...
+    return getattr(getattr(import_module(module_path), cls_name), attr)
+
+
+# Captured at conftest import, before any session fixture starts patching, so these
+# are the genuine constructors rather than a skip stub.
+_REAL_LLM_INITS = {dotted: _resolve(dotted) for dotted in LLM_MOCKED_METHODS}
+
+
+@pytest.fixture
+def allow_llm_config_construction():
```
cesarebernardis: This should not be done, please remove

pyagentspec/tests/adapters/conftest.py
```diff
@@ -112,6 +114,34 @@ def _skip(*_args, **_kwargs):
             p.stop()
 
 
+def _resolve(dotted: str) -> Any:
```
cesarebernardis: It's unclear what this method does, it needs a better name and at least a 1-line docstring. Inline imports are also very discouraged, avoid that unless strictly necessary (and motivate it in a comment)

pyagentspec/src/pyagentspec/managerworkers.py
```diff
@@ -65,6 +66,17 @@ class ManagerWorkers(AgenticComponent):
         default=AgentSpecVersionEnum.v25_4_2, init=False, exclude=True
     )
 
+    def _get_inferred_inputs(self) -> List[Property]:
+        # The group manager drives the conversation and is the component whose prompt
+        # the runtime renders, so the group accepts exactly the inputs the manager
+        # accepts. The hasattr guard matches Flow._get_inferred_inputs: validators can
+        # run this against a partially-constructed model with no group_manager yet.
+        return (self.group_manager.inputs or []) if hasattr(self, "group_manager") else []
```
cesarebernardis: This is not part of the contract of the `ManagerWorker` component in Agent Spec at the moment, so right now this is not respecting the language spec.

I understand why this is being done though, and it looks reasonable, so maybe we should add this to the language spec. We'll discuss internally and I'll come back to you on the topic.

pyagentspec/src/pyagentspec/adapters/langgraph/_managerworkers.py
```diff
@@ -0,0 +1,214 @@
...
+from pyagentspec.tracing.spans import (
+    ManagerWorkersExecutionSpan as AgentSpecManagerWorkersExecutionSpan,
+)
+
+# Cannot collide with a normalized worker node name, which is always [a-z0-9_].
+_MANAGER_NODE_KEY = "__manager__"
+
+#: Prefix of the synthetic ``delegate_to_<worker>`` tool names the manager's LLM uses
+#: to address a worker. Public so consumers can recognize the protocol.
+DELEGATE_TOOL_PREFIX = "delegate_to_"
```
cesarebernardis: This is not so uncommon as a name I think, so there's a chance that it will collide with the name of an actual tool, and you will try to reroute to a non-existing agent. I think you could use a more uncommon prefix (e.g., adding another underscore or 2, like the delegation keys below), and/or checking that its suffix is an actual worker node when doing delegation.

pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
```diff
@@ -554,6 +604,17 @@ def _format_agent_result(self, result: Dict[str, Any]) -> ExecuteOutput:
             ]
             return {}, NodeExecutionDetails(generated_messages=generated_messages)
 
+        if isinstance(self.node.agent, AgentSpecManagerWorkers):
+            # The hierarchical graph runs over MessagesState, which cannot carry a
+            # structured_response outward: the manager's final message is the result.
```
cesarebernardis: This is an adapter limitation, and I think that it's ok for now to have it, but we should probably fail early during conversion, when we see that a non-supported output shape is required.

pyagentspec/tests/adapters/langgraph/flows/test_managerworkers_node.py
```diff
@@ -0,0 +1,140 @@
+# Copyright © 2026 Oracle and/or its affiliates.
```
cesarebernardis: This file is inside `flows`, but it's not a Node. I think you can move everything in the main `test_managerworkers.py`

pyagentspec/src/pyagentspec/managerworkers.py
cesarebernardis: We discussed, and we think it makes sense to introduce this in the language specification. For simplicity, we can just say that the I/Os of the ManagerWorker must be the same I/Os of its manager (same name and type).

cesarebernardis:
@spichen Thank you for addressing my comments.

The PR is now ready, but we are in the middle of the release process, we will merge it right after completing the release.

dhilloulinoracle:
Internal regression failed: Build ID #563

dhilloulinoracle:
Internal regression failed: Build ID #564

dhilloulinoracle:
Internal regression failed: Build ID #565

dhilloulinoracle:
Merge Gate failed: Build ID #735

cesarebernardis (approved):

dhilloulinoracle (approved):

sonleoracle (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #512

dhilloulinoracle:
Merge Gate failed: Build ID #736

dhilloulinoracle:
Merge Gate failed: Build ID #737

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #513

dhilloulinoracle:
Merge Gate failed: Build ID #738

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #514

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #739

PR #212, opened 2026-07-27
RFC-DbmsVectorChainLlmConfig

tan-mayhem:
Proposes adding a built-in `DbmsVectorChainLlmConfig` component to PyAgentSpec.

This component would allow the official serializer to produce the LLM configuration accepted by Oracle Database's `DBMS_AI_AGENT` executor, including database-managed credential references, without requiring users to manually modify the serialized Agent Spec.

This PR contains the RFC proposal only; it does not include the implementation.


cesarebernardis:
Merged in https://github.com/oracle/agent-spec/pull/214

PR #234, opened 2026-08-21
fix(langgraph): add description to OCI structured-output schema

### LINKED ISSUE #232: LangGraph adapter fails to load OciGenAiConfig LlmNode with structured output
The LangGraph Agent Spec adapter fails while loading a valid Flow containing an `LlmNode` configured with `OciGenAiConfig` and a strict nested output schema.

Agent Spec serialization succeeds, but runtime loading fails before any OCI request is sent.

## Reproduction

The `LlmNode` declares a structured output such as:

```json
{
  "title": "result",
  "type": "object",
  "required": ["name", "profile"],
  "properties": {
    "name": {"type": "string"},
    "profile": {
      "type": "object",
      "required": ["score", "active", "tags"]
    }
  }
}
```

The LangGraph runtime is loaded with:

```text
RUNTIME_CLASS_IMPORT_PATH=langgraphruntime.LanggraphAgentSpecLoader
```

## Actual behavior

Loading the component fails with:

```text
ValueError: Unsupported tool type <class 'dict'>.
Tool must be passed in as a BaseTool instance, TypedDict class, or BaseModel type.
```

The failure occurs in:

```text
pyagentspec/adapters/langgraph/_node_execution.py
self.llm.with_structured_output(json_schema)
```

Then:

```text
ChatOCIGenAI.bind_tools([json_schema])
```

The adapter passes the JSON schema as a raw Python dictionary, but the installed `langchain_oci` provider rejects dictionary tool definitions.

## Expected behavior

The LangGraph adapter should either:

1. Convert the Agent Spec JSON schema into a supported structured-output representation before calling `ChatOCIGenAI`; or
2. Raise a clear unsupported-feature error during configuration validation.

The component should not fail with a low-level `Unsupported tool type <class 'dict'>` error.

## Impact

- Agent Spec serialization succeeds.
- LangGraph component loading fails.
- The OCI endpoint is never called.
- Structured LlmNode tests cannot execute with `OciGenAiConfig`.
- The same Agent Spec may behave differently across LangGraph and WayFlow runtimes.

AmirF194:
## Root cause

`LlmNodeExecutor.__init__` (`_node_execution.py`) builds the LangGraph structured-output
schema for an `LlmNode` as a bare dict with only `title`, `type` and `properties`, then
passes it to `self.llm.with_structured_output(json_schema)`.

For an `OciGenAiConfig`-backed model, that call reaches `langchain_oci`'s
`ChatOCIGenAI.bind_tools` -> `convert_to_oci_tool`. The generic provider (Meta/OpenAI/Gemini
model ids) only requires `title` and `properties`, so it happens to accept this dict. The
Cohere provider (`cohere.*` model ids) additionally requires `description`:

```python
if not all(k in tool for k in ("title", "description", "properties")):
    raise ValueError("Unsupported dict type. Tool must be a BaseTool instance, "
                      "JSON schema dict, or Pydantic model.")
```

so any Cohere-family model configured with a structured (non-single-string) `LlmNode`
output fails at executor construction time, before any request reaches OCI, matching the
report on #232.

## Fix

Add a static `description` key to the schema dict, mirroring the existing `title` comment.

## Verification

- Reproduced against current HEAD in a clean `python:3.12-slim` container (`pyagentspec`
  installed editable from this checkout, provenance checked via `module.__file__`):
  constructing `LlmNodeExecutor` for an `OciGenAiConfig(model_id="cohere.command-a-...")`
  node with two non-string outputs raises the `ValueError` above; a Meta/generic model id
  does not, which is why this is provider-specific rather than universal.
- Added `test_llmnodeexecutor_structured_output_supported_across_providers`
  (`test_ocigenai_conversion.py`), parametrized over a Cohere and a Meta model id, using a
  throwaway locally-generated OCI API key so it needs no real account or network call. It
  fails on `main` for the Cohere case and passes on this branch for both, run both ways in
  the same container.
- `black`, `isort`, `flake8 --select C801` (including the copyright check) and `bandit -c
  bandit.yaml -r pyagentspec/` all clean on the two changed files. `mypy` is not run here:
  the CI job excludes `pyagentspec/src/pyagentspec/adapters`.
- Full `pytest tests` (both changed and unchanged) with `SKIP_LLM_TESTS=1`, matching this
  repo's CI env: same 10 failed / 1042 passed / 666 skipped / 26 errors before and after this
  diff, all pre-existing and traced to optional extras not installed in this container
  (`langchain_ollama`, `langgraph_swarm`), not to this change.
- Not verified: like every other test in this file that constructs an `OciGenAiConfig`
  (including the pre-existing `test_reverse_convert_chatocigenai_to_agentspec`), the new test
  is skipped under the project's standard `SKIP_LLM_TESTS=1` CI run; I could only exercise it
  locally with that flag unset.

Fixes #232

### DIFF
--- a/pyagentspec/tests/adapters/langgraph/llms/test_ocigenai_conversion.py
+++ b/pyagentspec/tests/adapters/langgraph/llms/test_ocigenai_conversion.py
@@ -267,82 +267,51 @@ def test_reverse_convert_chatocigenai_to_agentspec_real():
     assert isinstance(client_cfg, OciClientConfigWithApiKey)
 
 
-def _write_throwaway_oci_config(directory: Path) -> str:
-    """Write a syntactically valid, locally-generated OCI API-key config/keypair.
-
-    Not a real credential and never sent anywhere: `ChatOCIGenAI` parses this file
-    to construct its client, but the LangGraph adapter code under test never issues
-    a network call, so no real OCI account is required.
-    """
-    from cryptography.hazmat.primitives import serialization
-    from cryptography.hazmat.primitives.asymmetric import rsa
-
-    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
-    key_path = directory / "oci_api_key.pem"
-    key_path.write_bytes(
-        key.private_bytes(
-            encoding=serialization.Encoding.PEM,
-            format=serialization.PrivateFormat.TraditionalOpenSSL,
-            encryption_algorithm=serialization.NoEncryption(),
-        )
-    )
-    config_path = directory / "config"
-    config_path.write_text(
-        "[DEFAULT]\n"
-        "user=ocid1.user.oc1..aaaaaaaathrowaway\n"
-        "fingerprint=aa:bb:cc:dd:ee:ff:00:11:22:33:44:55:66:77:88:99\n"
-        "tenancy=ocid1.tenancy.oc1..aaaaaaaathrowaway\n"
-        "region=us-ashburn-1\n"
-        f"key_file={key_path}\n"
-    )
-    return str(config_path)
-
-
-@pytest.mark.parametrize(
-    "model_id",
-    [
-        "cohere.command-a-03-2025",
-        "meta.llama-3.3-70b-instruct",
-    ],
-)
-def test_llmnodeexecutor_structured_output_supported_across_providers(
-    tmp_path: Path, model_id: str
-) -> None:
-    """`LlmNodeExecutor` must build a `with_structured_output` schema every
+def test_llmnodeexecutor_structured_output_schema_has_description() -> None:
+    """`LlmNodeExecutor` must pass `with_structured_output` a schema every
     `ChatOCIGenAI` provider backend accepts, not only the ones whose tool-schema
-    validation happens to tolerate a dict with no "description" key.
+    validation happens to tolerate a dict with no "description" key
+    (`langchain_oci`'s `CohereProvider.convert_to_oci_tool` requires one).
 
-    `ChatOCIGenAI` picks its provider (Cohere vs. Meta/generic) from `model_id`
-    alone (`AgentSpecToLangGraphConverter` never forwards `OciGenAiConfig.provider`),
-    so the two ids above exercise the two `convert_to_oci_tool` implementations.
+    Monkey-patches `_node_execution.BaseChatModel` with a tiny fake that
+    captures the schema `with_structured_output` is called with, instead of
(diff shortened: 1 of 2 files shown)


pyagentspec/tests/adapters/langgraph/llms/test_ocigenai_conversion.py
```diff
@@ -265,3 +265,84 @@ def test_reverse_convert_chatocigenai_to_agentspec_real():
...
+@pytest.mark.parametrize(
+    "model_id",
+    [
+        "cohere.command-a-03-2025",
+        "meta.llama-3.3-70b-instruct",
+    ],
+)
+def test_llmnodeexecutor_structured_output_supported_across_providers(
+    tmp_path: Path, model_id: str
+) -> None:
```
cesarebernardis: Thank you for the contribution!

This test is not really checking the new change, so I'd ask to replace it with another unit test near LlmNodeExecutor that:
- builds an `LlmNodeExecutor`;
- monkey-patches `_node_execution.BaseChatModel` with a tiny fake that exposes an ad-hoc `with_structured_output`;
- asserts the captured schema has the expected description and output properties.

This also does not require to have a proper OCI configuration, which would probably be ideal to write a proper e2e test, but unfortunately it's not currently feasible.

AmirF194:
Pushed 3694bb6: replaced the parametrized OCI test with a unit test near LlmNodeExecutor that monkeypatches `_node_execution.BaseChatModel` with a fake capturing the `with_structured_output` schema, and asserts its description and properties directly. No OCI config needed anymore.

cesarebernardis (approved):

cesarebernardis:
> Pushed [3694bb6](https://github.com/oracle/agent-spec/commit/3694bb684652454dea69392a0d523e0e8d150c21): replaced the parametrized OCI test with a unit test near LlmNodeExecutor that monkeypatches `_node_execution.BaseChatModel` with a fake capturing the `with_structured_output` schema, and asserts its description and properties directly. No OCI config needed anymore.

LGTM.

To proceed with the next steps to merge this PR, I would need you to sign the Oracle Contributor Agreement (OCA). Please check the comment at https://github.com/oracle/agent-spec/pull/234#issuecomment-5367389849

AmirF194:
Closing this one. I've decided not to sign the Oracle Contributor Agreement, so it can't go any further, and it isn't fair to leave it sitting in your queue.

Thanks for the review time, the fix itself is small if anyone wants to pick it up: the OCI structured-output schema needs a description field or the LangGraph path drops it.

PR #240, opened 2026-08-27
Agent Spec Evaluation Computer improvements

soufianeamini:
- Use CapacityLimiter over Semaphore
- Limit the amount of concurrent tasks that are running at once (the number of tasks spawned would be at most max_concurrency + 1)
- Producer/worker pattern is not needed anymore as a result

### DIFF
--- a/pyagentspec/tests/evaluation/_computers/test_evaluator_concurrency.py
+++ b/pyagentspec/tests/evaluation/_computers/test_evaluator_concurrency.py
@@ -97,31 +97,6 @@ async def test_sequential_running() -> None:
 
 
 @pytest.mark.anyio
-async def test_unlimited_concurrency() -> None:
-    num_samples = 200
-    dataset = Dataset.from_dict([{"dummy_arg": i} for i in range(num_samples)])
-    log_registry = LogRegistry()
-    callables: Dict[str, Callable[..., Awaitable[Any]]] = {
-        "dummy_callable": IoIntensiveMetric(log_registry, 1, 10, 1000)
-    }
-    computer = _AsyncCallablesComputer(
-        dataset=dataset,
-        callables=callables,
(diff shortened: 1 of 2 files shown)


pyagentspec/tests/evaluation/_computers/test_evaluator_concurrency.py
soufianeamini, line 1: @cesarebernardis The `test_unlimited_concurrency` test fails if the tasks finish too soon (if I add a bigger delay it succeeds) because it is not required anymore that all tasks start before starting processing. So from testing instead of seeing tasks go up to 200 running tasks concurrently, it ends up hovering around the 60 concurrent tasks range.

Is there a different kind of test you'd want to be implemented or should I just delete it?

cesarebernardis: No, I think you can get rid of it.

I am a bit more concerned by the removal of `test_run_does_not_spawn_one_task_per_item`, I think it's a good thing to check. What's the reason why it was removed? 

Maybe it does not apply anymore exactly as-is, as we are not using `start_soon`, but I would like to keep a semantically similar test if possible. Do we have a concurrency limit test already?

cesarebernardis (approved):

pyagentspec/tests/evaluation/_computers/test_evaluator_concurrency.py
soufianeamini: The reason for its removal is that the new version does actually spawn one task per item. It just doesn't do it in an unbounded fashion but rather respects the max_concurrency limit.

Yes we do have a test called `test_max_concurrency_is_respected`, if I (wrongly) increment the max_concurrency inside the CapacityLimiter constructor the test easily fails.

PR #258, opened 2026-09-11
fix(property): do not mutate schemas when normalizing union types

### LINKED ISSUE #254: Property type comparison mutates schemas that combine a `type` list with `anyOf`
## Description

Comparing two property schemas, or checking whether one is castable to the other, silently **mutates** a schema that combines a `type` list with `anyOf`: the entries of `type` are appended to the schema's own `anyOf` list. The schema keeps growing on every comparison.

## Reproduction

```python
from pyagentspec.property import json_schemas_have_same_type

schema_a = {"type": ["integer", "null"], "anyOf": [{"type": "string"}]}
schema_b = {"anyOf": [{"type": "string"}, {"type": "integer"}, {"type": "null"}]}

json_schemas_have_same_type(schema_a, schema_b)
print(schema_a)
# {'type': ['integer', 'null'], 'anyOf': [{'type': 'string'}, {'type': 'integer'}, {'type': 'null'}]}
#                                        ^ two entries were appended to the caller's schema

json_schemas_have_same_type(schema_a, schema_b)
print(len(schema_a["anyOf"]))   # 5
```

The same happens through `json_schema_is_castable_to` and `value_is_of_compatible_type`, which are called during `Flow` / `DataFlowEdge` validation, so a `Property.json_schema` can change as a side effect of building a flow.

## Expected behavior

Type comparison is read-only; the input schemas are left untouched.

## Root cause

`_normalize_json_schema_union_types` in `pyagentspec/src/pyagentspec/property.py` uses `all_types = schema.get("anyOf", [])` as the accumulator of normalized types and then appends to it, so when the schema already has an `anyOf` list the appends go into the caller's list.

## Environment

- pyagentspec `main` (26.4.0.dev0), Python 3.12

fede-kamel:
Fixes #254.

## Root cause

`_normalize_json_schema_union_types` in `property.py` used `schema.get("anyOf", [])` as the accumulator of normalized types and appended the entries of the `type` list to it. When the schema already had an `anyOf` list, the appends went into the caller's list, so every `json_schemas_have_same_type` / `json_schema_is_castable_to` / `value_is_of_compatible_type` call on such a schema grew its `anyOf`. These functions run during `Flow` and `DataFlowEdge` validation, so a `Property.json_schema` could change as a side effect of building a flow.

## Changes

- `property.py`: accumulate the normalized types in a copy of the `anyOf` list.
- `tests/test_properties.py`: regression test comparing a schema that combines a `type` list with `anyOf` several times and asserting both schemas are unchanged. Fails on `main`, passes with the fix.

## Verification

- `pytest tests/test_properties.py`: 177 passed, 2 skipped. Full suite: 1180 passed, 586 skipped (`SKIP_LLM_TESTS=1`).

### DIFF
--- a/docs/pyagentspec/source/changelog.rst
+++ b/docs/pyagentspec/source/changelog.rst
@@ -36,6 +36,12 @@ Improvements
 Bug fixes
 ^^^^^^^^^
 
+* **Property type comparison no longer mutates schemas**
+
+  Comparing or checking the castability of a JSON schema that combines a ``type`` list with
+  ``anyOf`` appended the normalized types to the schema's own ``anyOf`` list. The schema is now
+  left untouched.
+
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
--- a/pyagentspec/src/pyagentspec/property.py
+++ b/pyagentspec/src/pyagentspec/property.py
@@ -339,7 +339,8 @@ def _normalize_json_schema_union_types(schema: JsonSchemaValue) -> List[JsonSche
     json_schema_types = (
         json_schema_type if isinstance(json_schema_type, list) else [json_schema_type]
     )
-    all_types: List[JsonSchemaValue] = schema.get("anyOf", [])
+    # Copy the list: the normalized types must not be appended to the schema's own anyOf
+    all_types: List[JsonSchemaValue] = list(schema.get("anyOf", []))
     for json_schema_type in json_schema_types:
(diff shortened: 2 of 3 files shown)


cesarebernardis (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #580

sonleoracle (approved):

PR #260, opened 2026-09-11
fix(langgraph): recover agent outputs from the final message when no structured response exists

### LINKED ISSUE #224: LangGraph AgentNode returns tool-output field names instead of values
### Description

The LangGraph adapter invokes the AgentNode tool successfully, but the final AgentNode output contains the tool-result field names instead of the returned data values.

The failure is observed in:

- `tests/flows/nodes/agentnodes/validation/test_valid_configs_agentnode.py`
  - `test_agentnode_with_tool_can_be_executed`

The test tool returns the following values:

```python
"Alice"  -> ["Alice", "No", "USA"]
"Karim"  -> ["Karim", "Yes", "Morocco"]
"Khalid" -> ["Khalid", "Unknown", "Unknown"]
```

The CTS log confirms that the tool is called, but LangGraph returns:

```text
["name", "CEO", "country"]
```

instead of the expected values.

The failure is therefore not caused by missing tool registration or failure to invoke the tool. It occurs after tool execution, while the result is being converted into the AgentNode output.

`AgentNodeExecutor._format_agent_result()` delegates output extraction to `extract_outputs_from_invoke_result()`:

```python
outputs = extract_outputs_from_invoke_result(
    result,
    self.node.outputs or [],
)
```

The extractor gives priority to the complete `structured_response` dictionary and then copies matching output fields from the main result dictionary. It does not validate that the extracted value corresponds to the actual tool result or that the structured response contains the expected value type.

The current behavior suggests that the structured response or model-generated output is being populated with the tool schema/default field names rather than the result returned by `get_person_info()`.

### Expected behavior

The AgentNode should return:

```python
{"search_results": ["Alice", "No", "USA"]}
```

for Alice, with equivalent value arrays for Karim and Khalid.

The adapter should preserve the actual tool return value through:

1. Tool invocation.
2. LangGraph state update.
3. Structured response generation.
4. AgentNode output extraction.
5. CTS result mapping.

### Relevant code

- `agent-spec/conformance_test_suite/tests/flows/nodes/agentnodes/validation/test_valid_configs_agentnode.py`
- `agent-spec/pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py:549-558`
- `agent-spec/pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py:922-942`
- `agent-spec/pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py:1143-1193`
- `agent-spec/conformance_test_suite/runtimes/langgraphruntime/src/langgraphruntime/runtime.py:119-162`

fede-kamel:
Fixes #224.

## Root cause

The `["name", "CEO", "country"]` returned by the CTS test is not the tool's field names leaking through: it is the **declared default** of the `search_results` output in the test's Agent Spec. LangChain's `create_agent(response_format=ToolStrategy(...))` only sets `structured_response` when the model calls the structured-output tool; when the model answers in plain text instead, the run ends with no structured response and `extract_outputs_from_invoke_result` fell through to the declared defaults. `AgentNodeExecutor` also seeded `structured_response: {}`, which the extractor treated as a present-but-empty response.

The CTS deterministic model answers the second turn with the tool result as plain text (it only understands provider-native `response_format`), which is why Wayflow passes and LangGraph returns the default. Reproduced live against OCI Generative AI models as well (a model answering in prose after the tool call produces the same default), and deterministically with a fake chat model.

## Changes

- `langgraph/_node_execution.py`: `extract_outputs_from_invoke_result` treats a missing or empty `structured_response` as absent and recovers the outputs from the final AI message when it is compatible with the declared schema: a JSON object keyed by output names, a JSON value for a single non-string output, or the text itself for a single string output (text content blocks are joined). Precedence is defaults, then recovered values, then `structured_response`, then values already present in the state. A `logging` warning names the outputs that fell back to their defaults. Final messages carrying tool calls are never used.
- `tests/adapters/langgraph/flows/test_agentnode_output_recovery.py`: the CTS scenario for the three names with a fake chat model, the unchanged structured-response path, the default fallback with its warning, and unit cases of the extractor. Eight tests fail on `main`, thirteen pass with the fix. No LLM calls.

## Verification

- `SKIP_LLM_TESTS=1 pytest tests/adapters/langgraph`: 163 passed, 89 skipped.

## Notes for reviewers

- PR #210 adds a `StructuredOutputGuard` that raises when `structured_response` is absent; if it lands first, the guard should run after (or skip when) this recovery succeeds.
- The schema-fidelity PR also edits `extract_outputs_from_invoke_result` (wraps the result in `to_json_value`); whichever lands second needs a trivial rebase.
- Two related observations were left out of scope: some models emit the structured-output call in the same turn as the tool call under `tool_choice="any"`, and Cohere models on OCI reject `tool_choice` entirely, so `ToolStrategy` fails for them at load time.

### DIFF
--- a/docs/pyagentspec/source/changelog.rst
+++ b/docs/pyagentspec/source/changelog.rst
@@ -36,6 +36,15 @@ Improvements
 Bug fixes
 ^^^^^^^^^
 
+* **LangGraph adapter: agent outputs answered as plain text**
+
+  When an agent with declared outputs ended its run with a plain message instead of calling
+  the structured output tool, the adapter silently reported the declared defaults of the
+  outputs (for example the field names used as default of an ``AgentNode`` output). The
+  values are now recovered from the final agent message when it can be mapped to the
+  declared outputs (a JSON object keyed by output name, the JSON value of a single output,
+  or the free text of a single string output), and a warning is logged when they cannot.
+
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
--- a/pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
+++ b/pyagentspec/src/pyagentspec/adapters/langgraph/_node_execution.py
@@ -50,6 +50,7 @@ from pyagentspec.flows.nodes import StartNode as AgentSpecStartNode
 from pyagentspec.flows.nodes import ToolNode as AgentSpecToolNode
 from pyagentspec.property import Property as AgentSpecProperty
 from pyagentspec.property import _empty_default as pyagentspec_empty_default
+from pyagentspec.property import value_is_of_compatible_type
 from pyagentspec.tracing.events import NodeExecutionEnd as AgentSpecNodeExecutionEnd
 from pyagentspec.tracing.events import NodeExecutionStart as AgentSpecNodeExecutionStart
 from pyagentspec.tracing.events.exception import ExceptionRaised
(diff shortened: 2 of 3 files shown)


cesarebernardis (approved):

cesarebernardis (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #579

Issue #227, opened 2026-08-14
LangGraph agent executions reach recursion limit across basic and provider-backed configurations


JainSajal23:
### Description

Multiple LangGraph agent execution tests load successfully but do not terminate. Instead, execution repeatedly traverses the graph until LangGraph raises:

```text
langgraph.errors.GraphRecursionError:
Recursion limit of 800 reached without hitting a stop condition.
```

The affected test groups are:

- Basic agent execution.
- VLLM execution.
- OpenAI execution.
- OpenAI-compatible execution.

Representative tests include:

- `tests/agent/validation/test_valid_configs_agent.py`
- `tests/llms/execution/test_execution_llmconfig_with_agent.py`

The affected LLM configurations include:

- `vllmconfig_with_agent.yaml`
- `openaiconfig_with_agent.yaml`
- `openaicompatibleconfig_with_agent.yaml`

The LangGraph CTS runtime explicitly sets:

```python
{"configurable": {"thread_id": "1"}, "recursion_limit": 800}
```

in `langgraphruntime/runtime.py:101-103`.

The logs show that the graph reaches this limit without reaching a terminal state. This occurs across multiple model-provider paths, which suggests a shared graph-construction, response-normalization, or termination issue rather than an isolated provider failure.

The LangGraph adapter builds the agent graph through `_create_react_agent_with_given_info()` and configures the graph with:

- The converted model.
- Converted tools.
- The Agent Spec system prompt.
- The Agent Spec output model.
- The LangGraph state schema.
- The configured checkpointer.

Relevant construction code is in `_langgraphconverter.py:1143-1193`.

Possible failure modes include:

- A normal assistant response is incorrectly interpreted as an incomplete continuation.
- A model response is repeatedly classified as a tool call.
- The graph routing condition does not recognize the terminal response.
- The response format or structured-response extraction prevents the stop condition.
- The provider-specific response shape is not normalized into the message format expected by the graph.
- A tool or model node returns state that causes the graph to re-enter the same node.

The Ollama execution failure should be tracked separately unless the endpoint is first made available, because its log reports connection refusal rather than recursion. OCI cases should also be handled separately because they fail during authentication/configuration setup.

### Expected behavior

For a valid model endpoint and valid credentials, a basic request such as:

```text
calculate 3x6 ?
```

should execute once, reach a terminal assistant response, and return a valid CTS execution status containing either:

- Non-empty `agent_messages`, or
- Non-empty `outputs`.

[...]

Issue #223, opened 2026-08-13
LangGraph client-tool confirmation interrupt raises KeyError: 'name'


JainSajal23:
### Description

The LangGraph runtime fails while converting a client-tool confirmation interrupt into the CTS `ToolRequest` format.

The failure is observed in:

- `tests/clienttools/execution/test_agent_execution_with_1_clienttool_confirmation.py`
  - `test_agent_with_1_clienttool_confirmation_accept_flow`
  - `test_agent_with_1_clienttool_confirmation_reject_flow`

The LangGraph adapter creates the confirmation payload in `_langgraphconverter.py` using the following structure:

```python
{
    "action_requests": [
        {
            "name": tool_name,
            "arguments": tool_arguments,
            "description": ...
        }
    ],
    "review_configs": [...]
}
```

Therefore, the tool name is nested under `action_requests[0]["name"]`; there is no top-level `name` field.

However, `langgraphruntime/runtime.py` treats every interrupt as a direct tool request and executes:

```python
tool_request = interrupt.value
ToolRequest(
    name=tool_request["name"],
    ...
)
```

This occurs at `runtime.py:139-144` and raises:

```text
KeyError: 'name'
```

The runtime therefore fails before the CTS test can complete the confirmation accept or reject flow. The confirmation interrupt is generated correctly, but it is interpreted using the wrong interrupt schema.

The same runtime path is also used for client-tool execution, which explains why both client-tool confirmation cases fail with the same error.

### Expected behavior

The runtime should distinguish between:

- A confirmation interrupt containing `action_requests` and `review_configs`.
- A direct client-tool request containing top-level `name`, `description`, and `inputs`.

A confirmation interrupt should be converted into the CTS confirmation status, allowing the caller to invoke `confirm_or_reject_tool_confirmation()` and continue with either the accepted or rejected path.

### Relevant code

- `agent-spec/conformance_test_suite/runtimes/langgraphruntime/src/langgraphruntime/runtime.py:135-150`
- `agent-spec/pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py:950-970`
- `agent-spec/pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py:1784-1806`

fede-kamel:
[...]

PR #261, opened 2026-09-11
fix(langgraph): pass the recursion limit of the flow run on to AgentNode agents

fede-kamel:
Relates to #227.

## Root cause

Two things combine in the CTS failures reported in #227:

1. The non-termination itself: the adapter enables `response_format=ToolStrategy(...)` for any agent with declared outputs. With no tools, LangChain's `create_agent` routes from the model back to the model until a `structured_response` exists, so a model that answers in prose (the CTS deterministic LLM server, or any provider ignoring the structured-output tool) loops until `GraphRecursionError`. Reproduced deterministically with a fake chat model (agent with outputs and no tools: recursion error after 12 model calls; same agent without outputs: terminates in 2 calls). Open PR #210 addresses this part and, applied locally, makes the CTS case terminate in 2 calls.
2. **What this PR fixes**: `create_agent` binds `recursion_limit=9999` to the agents it compiles, and `AgentNodeExecutor` invoked the nested agent with the loader config only. The limit configured for the enclosing flow run therefore never applied inside an `AgentNode`: with `recursion_limit=12` on the flow run, a non-terminating nested agent made 9999 model calls before failing.

## Changes

- `langgraph/_node_execution.py`: `AgentNodeExecutor` reads the config of the current run and passes an explicitly configured recursion limit (any value other than LangChain's default of 25) on to the nested agent. A limit set in the loader config still takes precedence, and runs using the default limit keep the agent's own limit so tool-heavy agents are not cut at 25 steps.
- `tests/adapters/langgraph/flows/test_agentnode_recursion_limit.py`: sync and async runs with `recursion_limit=6` stop the nested agent after 3 model calls (5000 calls on `main` before the test was capped), the loader config wins over the run config, and a run with the default limit keeps the agent's limit. Fake chat model, no LLM calls.

## Verification

- `SKIP_LLM_TESTS=1 pytest tests/adapters/langgraph`: 113 passed, 89 skipped.
- Live check against OCI Generative AI models: a compliant model terminates in one turn with the structured-output tool call, so the loop only shows with models or servers that answer in prose.

## Notes for reviewers

- Independent of #210; both can land in any order.
- The CTS runtime in #117 also treats every interrupt as a client-tool request, which is the `KeyError: 'name'` of #223; a proposed patch is posted on that issue.

### DIFF
--- a/docs/pyagentspec/source/changelog.rst
+++ b/docs/pyagentspec/source/changelog.rst
@@ -36,6 +36,14 @@ Improvements
 Bug fixes
 ^^^^^^^^^
 
+* **LangGraph adapter: AgentNode agents honour the recursion limit of the flow run**
+
+  The agents compiled for Flow ``AgentNode`` steps are bound by LangChain to a recursion
+  limit of 9999 steps, which overrode the ``recursion_limit`` configured for the flow run.
+  An agent that never terminated made thousands of model calls before failing. A recursion
+  limit explicitly configured for the run (or in the loader config) is now passed on to the
+  nested agent; runs using LangChain's default limit keep the agent's own limit.
+
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


cesarebernardis (approved):

cesarebernardis (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #578

sonleoracle (approved):

PR #264, opened 2026-09-14
feat(adapters): run OCI Generative AI models in AutoGen, Agent Framework and OpenAI Agents through the OpenAI-compatible API

### LINKED ISSUE #263: Run OciGenAiConfig models in the AutoGen, Agent Framework and OpenAI Agents adapters (OpenAI-compatible API)
## Description

`OciGenAiConfig` is only executable by the WayFlow and LangGraph adapters. The adapters built on the `openai` SDK (AutoGen, Microsoft Agent Framework, OpenAI Agents) reject it:

```text
NotImplementedError: LlmConfig with api_provider='oci' is not supported in autogen yet. Consider using a specific LlmConfig subclass instead.
```

(same message in `agent_framework` and `openaiagents`), so an Agent Spec configuration that uses OCI Generative AI is not portable across the supported runtimes.

## Proposal

OCI Generative AI exposes OpenAI-compatible `chat/completions` and `responses` endpoints under `<service_endpoint>/openai/v1`, authenticated with an OCI request signature and an `opc-compartment-id` header. The three adapters can execute `OciGenAiConfig` by giving their OpenAI clients a signing `httpx` client built from the `OciClientConfig` (the four authentication types map onto the `httpx.Auth` implementations of the `oci-openai` package), with the compartment header, and by mapping `api_type` to chat completions or the Responses API where the runtime supports it. The runtime-to-Agent-Spec converters should rebuild the `OciGenAiConfig` from such clients.

Known limits of that API to handle explicitly: Cohere models and the `DEDICATED` serving mode are not served; the Responses API needs OpenAI models or a conversation store; the endpoint negotiates `zstd` and compresses streamed chunks as independent frames that `httpx` cannot decode (gzip works).

## Environment

- pyagentspec `main` (26.4.0.dev0), autogen-core 0.7.4, agent-framework-core 1.13.0, openai-agents 0.19.4

fede-kamel:
Closes #263.

Makes `OciGenAiConfig` executable by the three adapters built on the `openai` SDK: AutoGen, Microsoft Agent Framework and OpenAI Agents. Today they reject it with `api_provider='oci' is not supported`, so an OCI-backed Agent Spec only runs on WayFlow and LangGraph.

## Approach

OCI Generative AI exposes OpenAI-compatible `chat/completions` and `responses` endpoints under `<service_endpoint>/openai/v1`; requests need an OCI request signature and the compartment in the `opc-compartment-id` header. A new shared module, `adapters/_oci_openai_common.py`, builds a signed `openai` client from any `OciGenAiConfig`:

- the four `OciClientConfig` authentication types map onto the `httpx.Auth` implementations of Oracle's `oci-genai-auth` package (the GA successor of `oci-genai-auth`) (API key, session token, instance principal, resource principal), installed by the new `oci` extra (`pip install "pyagentspec[oci]"`); a missing package raises an actionable `ImportError`;
- the compartment and `conversation_store_id` headers are set on the client, `retry_policy.max_attempts` / `request_timeout` map to the client retries and timeout;
- `validate_oci_openai_compatible_config` fails early for what the OpenAI-compatible API does not serve (details below);
- `oci_genai_config_from_openai_client` rebuilds the `OciGenAiConfig` for the runtime-to-Agent-Spec converters.

Each adapter then adds an `OciGenAiConfig` branch before its bare-`LlmConfig` fallback:

| Adapter | Chat completions | Responses API | Reverse conversion |
|---|---|---|---|
| AutoGen | `OpenAIChatCompletionClient(http_client=<signing client>, base_url, model_info)` | rejected (AutoGen has no Responses client) | yes |
| Agent Framework | `OpenAIChatCompletionClient(async_client=...)` | `OpenAIChatClient(async_client=...)`; without a `conversation_store_id` the agent runs with `store=False` (see findings) | yes |
| OpenAI Agents | `OpenAIChatCompletionsModel(model, client)` | `OpenAIResponsesModel(model, client)` | yes; code generation rejects OCI configs with a clear `UnsupportedPatternError` (the generated code only carries the model name) |

The LangGraph and WayFlow adapters are unchanged: they use the native OCI API.

## Verified live against OCI Generative AI

Agent Spec `Agent` with a `multiply(a, b)` server tool, loaded through each adapter's `AgentSpecLoader`, prompt "What is 6 times 7? Use the tool." (session-token authentication):

| Adapter | Model | API | Result |
|---|---|---|---|
| AutoGen | openai.gpt-4.1 | chat completions | tool called, "6 times 7 is 42." |
| Agent Framework | openai.gpt-4.1 | chat completions | tool called, "6 times 7 is 42." |
| Agent Framework | openai.gpt-4.1 | responses | tool called, "6 times 7 is 42." |
| OpenAI Agents | openai.gpt-4.1 | chat completions | tool called, "6 times 7 is 42." |
| OpenAI Agents | openai.gpt-4.1 | responses | tool called, "6 times 7 is 42." |
| OpenAI Agents | meta.llama-3.3-70b-instruct | chat completions | tool called, "The result of 6 times 7 is 42." |

The shared client itself was also exercised directly: chat completions, tool calls, JSON-schema structured output, streaming and the async client on gpt-4.1 and llama-3.3-70b.

## Findings about the OpenAI-compatible API, and how they are handled

- **Streaming and zstd.** The endpoint negotiates `zstd` when the client advertises it and compresses each streamed chunk as an independent zstd frame; `httpx` fails with `cannot use a decompressobj multiple times`. The client sends `Accept-Encoding: gzip, deflate`; gzip streams decode correctly. Worth a report to the service.
- **Cohere models** are not served (`400 Unsupported OpenAI operation`); rejected at conversion with a message pointing to the native-API runtimes.
- **`DEDICATED` serving mode** is rejected at conversion (endpoint OCIDs are not accepted as model names by this API).
- **Responses API with non-OpenAI models** requires a conversation store header; documented. Without a conversation store the service runs in zero-data-retention mode and rejects `previous_response_id`, so the Agent Framework agent is created with `store=False` in that case.
- **Llama models** reject tool calling when the first message is a system prompt (`400 Tool calling for this Llama model requires the first message to be a user message`), which is how AutoGen and Agent Framework order messages; they also send tool arguments as strings. Documented as limitations, not worked around here.
- **Gemini** returns no tool calls through this endpoint even with `tool_choice="required"`. Documented.

## Tests

- `tests/adapters/test_oci_openai_common.py`: the signed request is captured through an `httpx.MockTransport` for API-key and session-token authentication (signature key id, `x-content-sha256`, compartment header, encoding, retries and timeout), validation of unsupported configurations, reverse conversion. The OCI configuration file with a generated key is a shared fixture in `tests/adapters/conftest.py`.
- `tests/adapters/{autogen,agent_framework,openaiagents}/test_oci_genai.py`: client class per `api_type`, signed end-to-end requests through each framework against the mock transport (`client.create`, `Agent.run` / `get_response`, `Runner.run`), rejections before any client is built, `store` handling, config round trips.
- `tests/adapters/test_oci_openai_live.py`: opt-in live checks (`OCI_OPENAI_TEST_PROFILE`, `OCI_OPENAI_TEST_COMPARTMENT_ID`).

Offline tests build configurations with `model_construct`, so they run under `SKIP_LLM_TESTS=1`; the round-trip tests construct `OciGenAiConfig` normally and are skipped there, like the existing OCI tests.

Full suite with the adapters installed: all passing (see the PR checks for the core matrix). Public CI steps (black, isort, flake8 + copyright, bandit, mypy, `tests/run_tests.sh`) reproduced locally on Python 3.10 and 3.14.

## Follow-ups (not in this PR)

- CrewAI: goes through LiteLLM's native `oci/` provider rather than the `openai` SDK; separate change.
- LangGraph: `api_type` `openai_*` could map to `langchain-oci`'s `ChatOCIOpenAI` instead of the native client.
- OpenAI Agents code generation for OCI configurations needs the sensitive `auth_file_location` in the IR path.


fede-kamel:
Switched the request-signing dependency from `oci-openai` (now labelled legacy by Oracle) to `oci-genai-auth` 1.1.1, its GA successor with the same `httpx.Auth` classes and no dependency on the `openai` SDK. The `oci` extra, constraints, tests and docs follow; no behaviour change (offline signed-request tests and the live checks pass).

fede-kamel:
Follow-up pushed (130c2f4): the shared helper now also accepts **OCI Generative AI API keys**, the `OciClientConfigWithGenAiApiKey` component proposed in #265 (implementation on `fede-kamel:feat/oci-genai-api-key-auth`).

For a client configuration with `auth_type` `GENAI_API_KEY`, `create_oci_openai_client` builds a plain bearer `openai` client (`Authorization: Bearer sk-...`): no request signature and no `oci-genai-auth` import, the key taken from the configuration or from the `OCI_GENAI_API_KEY` environment variable. The detection relies on the `auth_type` discriminator of the specification, so this branch works with or without #265 merged. The reverse conversion rebuilds the client configuration **without the key** (sensitive-field convention) when the component is available and returns `None` otherwise, in which case the adapters fall back to their generic handling. The compartment header is still sent with bearer requests: the endpoint accepts it, and it lets `oci_genai_config_from_openai_client` recover `compartment_id`.

Live-verified with a Generative AI API key on `meta.llama-3.3-70b-instruct` through the helper: chat completion, streaming and a tool call all succeeded, both with and without the compartment header.

Tests: `tests/adapters/test_oci_openai_common.py` gains bearer tests (sync and async requests captured through the mock transport asserting `Authorization: Bearer` and no `x-content-sha256`, the environment fallback, the rejection of a signing auth for API keys, and the reverse conversion without the key). The reverse-conversion assertion is conditional on the component being importable until #265 lands. Offline OCI adapter tests: 51 passed (47 passed, 4 skipped under `SKIP_LLM_TESTS=1`). The adapter docs mention the new option.

